PR #2, opened 2019-12-19
Some code cleanup

timyardley:
This has a variety of consistency code cleanups in it, some markup for fixme's and a changeover to clowderframework instead of clowder.ncsa for reference URLs


app/services/ToolManagerPlugin.scala
```diff
@@ -310,7 +310,7 @@ class ToolManagerPlugin(application: Application) extends Plugin {
       val apipath = play.Play.application().configuration().getString("toolmanagerURI") + "/instances/" + instance.toolPath
 
       val statusRequest: Future[Response] = url(apipath).put(Json.obj(
-        "dataset" -> (dsURL.replace("/datasets", "/api/datasets")+"/download"),
+        "dataset" -> (dsURL.replace("/datasets", "/api/datasets") + ß"/download"),
```
robkooper: not sure the ß is supposed to be there.

app/services/mongodb/MongoDBFileService.scala
```diff
@@ -909,21 +909,21 @@ class MongoDBFileService @Inject() (
...
+    Logger.debug("thequery: " + theQuery.toString)
     FileDAO.find(theQuery).toList
   }
 
   def searchUserMetadataFormulateQuery(requestedMetadataQuery: Any): List[File] = {
-    Logger.debug("top: "+ requestedMetadataQuery.asInstanceOf[java.util.LinkedHashMap[String,Any]].toString()  )
+    Logger.debug("top: " + requestedMetadataQuery.asInstanceOf[java.util.LinkedHashMap[String,Any]].toString()  )
     var theQuery =  searchMetadataFormulateQuery(requestedMetadataQuery.asInstanceOf[java.util.LinkedHashMap[String,Any]], "userMetadata")
-    Logger.debug("thequery: "+theQuery.toString)
+    Logger.debug("thequery: " + ßtheQuery.toString)
```
robkooper: Same here, looks liked there is an extra ß

app/services/ToolManagerPlugin.scala
timyardley: This is corrected, sorry about that. Probably was an accidental keystroke.

app/services/mongodb/MongoDBFileService.scala
timyardley: This is corrected, sorry about that. Probably was an accidental keystroke.

lmarini:
Hi @timyardley . Thank you for the contributions. A few comments:

1. We follow https://www.atlassian.com/git/tutorials/comparing-workflows/gitflow-workflow. Anything that is not a hotfix should be branched from `develop` and merged back into `develop`. I think this PR is not too far off the current `develop`, so we should be able to just point it to `develop`.
2. There is a lot of changes in this PR. I understand most of them are code styling changes. Could you please share the rules you used? While the scala templates are compiled and any error would be picked up at compile time, the javascript is not checked. As such we might be introduce runtime errors.
3. Was there anything else that was changed besides formatting, spelling and urls?
4. Please add an entry to the `CHANGELOG.md` and `CONTRIBUTORS.md`.
5. We currently have a lot of activity (PRs) happening on 
https://opensource.ncsa.illinois.edu/bitbucket/projects/CATS. Unfortunately bitbucket doesn't show PRs unless you are logged in. If you are interested in looking at what current PRs are against `develop` please signup following the signup link here https://opensource.ncsa.illinois.edu/confluence/#recently-worked.

Thank you!

app/Iterators/BagItIterator.scala
```diff
@@ -20,7 +20,7 @@ class BagItIterator(pathToFolder : String, collection : Option[models.Collection
 
   private def addBagItTextToZip(pathToFolder : String, totalbytes: Long, totalFiles: Long, zip: ZipOutputStream, collection: models.Collection, user: Option[models.User]) = {
     zip.putNextEntry(new ZipEntry(pathToFolder + "/bagit.txt"))
-    val softwareLine = "Bag-Software-Agent: clowder.ncsa.illinois.edu\n"
+    val softwareLine = "Bag-Software-Agent: clowder\n"
```
lmarini: I couldn't find any specific information about what should go in this field. Did you find something regarding this? I saw something along the lines emails: `clowder <https://clowderframework.org/>`.

timyardley:
[scalastyle_config.xml.txt](https://github.com/clowder-framework/clowder/files/4135071/scalastyle_config.xml.txt)

this is a generic scalastyle config file that was part of the basis used. not all issues were resolved.

app/Iterators/BagItIterator.scala
timyardley: the agent definition is typically the name of the software, along with versioning. For example, this is how a similar field is used in the HTTP standard:

https://en.wikipedia.org/wiki/User_agent

My reasoning here would be that at a minimum it should be clowder rather than a specific domain as a clowder instance can (and will) be installed in other locations than at NCSA

timyardley:
I prefer to not use bitbucket for public project pushes, so happy to provide this patch here on github. I don't really have the cycles to shift it over to the other repo. Moving it to develop branch shouldn't be much of a lift, but again up to you guys if you want to do it. I don't actively have the cycles to convert it over.

My commits in the history gives the high-level of what each grouping of changes were... all minor cleanups, with some of the "domain" specific removals decoupling from NCSA.

Future submissions I can direct to github/develop, your choice if you bring this in or reject it. I don't care about the contribution credit, as this is a trivial cleanup patch not anything worthy of note.

lmarini:
Merged most of the changes in master. Left out formatting changes. Thank you!

PR #19, opened 2020-07-01
Queue returns no actions if db inaccessible

max-zilla:
BUG: If Clowder instance was using ElasticsearchQueue enabled, a thread will check every 5ms for actions queued in MongoDB collection to perform. If MongoDB goes down (even briefly e.g. switchover of PRIMARY) and that thread tries to fetch an action, the thread would crash and the queue would never get processed, only grow.

This adds a try/catch around the fetching of next queued action, and if an error is encountered it simply returns none ("no actions to take.") It will continue to return none until Mongo is accessible again, at which point it resumes processing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## Unreleased
+
+### Fixed
+- Queue threads (e.g. indexer) will no longer crash permanently if the queue connection is lost temporarily.
+
 ## 1.10.0 - 2020-06-30
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
-## Unreleased
+## 1.10.1 - Unreleased
 
 ### Fixed
 - Queue threads (e.g. indexer) will no longer crash permanently if the queue connection is lost temporarily.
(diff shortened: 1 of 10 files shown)


lmarini:
Can we make this a hot fix against master?

max-zilla:
> Can we make this a hot fix against master?

Changed it to be against master, didn't see a hotfix label or other indicator.

app/services/mongodb/MongoDBQueueService.scala
```diff
@@ -68,7 +69,18 @@ trait MongoDBQueueService {
...
   // get next entry from queue
   def getNextQueuedAction(): Option[QueuedAction] = {
-    return Queue.findOne(new MongoDBObject)
+    try {
+      val response = Queue.findOne(new MongoDBObject)
+      response
+    } catch {
+      case e: MongoException => {
+        // TODO: Will generate an error message every 5ms while Mongo is inaccessible...
+        //Logger.error("Problem connecting to MongoDB queue.")
```
lmarini, line 78: Could we set the level to trace? For those cases when we do want to see the errors?

CHANGELOG.md
```diff
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## Unreleased
```
lmarini, line 7: Since this is a hot fix, can you up the version number to 1.10.1 everywhere:

- Changelog
- Built.scala
- swagger.yml
- conf.py

I usually just search for the prior version number, in this case 1.10.0.

max-zilla: updated in those places

app/services/mongodb/MongoDBQueueService.scala
max-zilla: kept as error, but with a new flag so it only prints once per broken connection.

PR #41, opened 2020-07-27
simple test sends the first time success notification.

bingzhang:
## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [x] When possible

## Types of changes

## Checklist:

- [ ] I have signed the CLA
- [x] I have updated the documentation accordingly.
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.


bingzhang:
@robkooper if we check the date, we still need to calculate the # failures or # successes. This also depends on how many times the script run per day. I think it is similar to what the current pull request does.

robkooper:
I was thinking something along, this will send a message on success after failure, and the first time after midnight.

```
if [ job fails ]; then
  send failure message
  remove memory file
else
  today = $(date +'%Y-%m-%d') 
  if [ ! -e memoryfile ]; then
     send success message
  else if [ "$(cat memoryfile)" != "$today"]; then
     send success message
  fi
  echo $today > memoryfile
fi
```

scripts/tester/tester.sh
```diff
@@ -83,15 +84,24 @@ while [ $FILE_EXTRACTED -eq 0 ]; do
...
 echo "File extraction complete."
-    
+
+today=$(date +'%Y-%m-%d')
+if [ ! -e memoryfile ]; then
+     post_message "Upload+extract test script success!"
+elif [ "$(cat memoryfile)" != "$today" ]; then
+     post_message "Upload+extract test script success!"
+fi
+
```
robkooper, line 101: I think you need to write the $today to `memoryfile`

lmarini: @robkooper is the fix from @bingzhang good? Can we merge this?

PR #45, opened 2020-08-14
Better support for colon characters in search

max-zilla:
This modifies how we parse search strings from API/search box, as well as how a search is built from the Metadata Search page.

- wrap search terms in escaped " characters so a colon character isn't misinterpreted by Elasticsearch
- break search box strings down without stripping : or other operators so they can exist inside field names and search terms

To test this, create a metadata.json file with contents like this (modify IDs as needed):
```
{
  "@context": [
    "https://clowderframework.org/contexts/metadata.jsonld",
    {
      "csv_headers": "http://clowder.ncsa.illinois.edu/metadata/ncsa.csvheaders"
    }
  ],
  "attached_to": {
    "resource_type": "cat:file",
    "url": "http://localhost:9000/files/5efb87e9066c08e3f6cc5046"
  },
  "created_at": "Wed Jun 20 11:40:25 CDT 2018",
  "agent": {
        "@type": "cat:user",
        "user_id": "http://localhost:9000/api/users/5efb83f962ebfe31bbba975d"
  },
  "content": {
    "Fish:contactz": "banchovy:223-456-7890",
    "syn:loc_selector": "5310:123-456-7890",
    "IMPORTANT NUMBER": 34
  }
}
```
Post this to a file:
`curl -X POST -H "Content-Type: application/json" -d @metadata.json http://localhost:9000/api/files/5efb87e9066c08e3f6cc5046/metadata.jsonld\?key\=r1ek3rs`

Assuming Elasticsearch is running properly, you should now be able to search in the search box:
`"syn:loc_selector":"5310:123-456-7890"`
..and find your result.

Incidentally due to the way ES tokenizes things, this would also work:
`"syn:loc_selector":5310`

...but this would not:
`"syn:loc_selector":531`


lmarini:
Missing changelog. Do we want this to be a hotfix? I thought it didn't need to be.

max-zilla:
> Do we want this to be a hotfix? I thought it didn't need to be.

Couldn't remember how industry project felt about it. I will change it to develop if you don't need it to be against master.

PR #46, opened 2020-08-14
Better support for elasticsearch characters

max-zilla:
Building off this hotfix: https://github.com/clowder-framework/clowder/pull/45

...this takes a step further to clean up the string parsing logic and enable some of the long-intended features in the search box, namely use of other characters than : (e.g. ==, <, <=, !=) for comparison operators. The < operators are intended for numerical fields only.

If you uploaded the example metadata from that PR, try the following searches:
"IMPORTANT NUMBER"<35
"IMPORTANT NUMBER">33
"IMPORTANT NUMBER"<=34
"IMPORTANT NUMBER"!=34 (this will return everything else)

Also improves documentation, but did NOT yet add to the Search Syntax help until we have a chance to discuss.


lmarini:
Changelog missing.

lmarini:
I am wondering if there is a way to make it so that "IMPORTANT NUMBER"!=34 only returns documents that have the field "IMPORTANT NUMBER" set but of which the value is not 34.

app/services/ElasticsearchPlugin.scala
```diff
@@ -837,73 +822,86 @@ class ElasticsearchPlugin(application: Application) extends Plugin {
...
-      *  ==  equals (exact match)
-      *  !=  not equals (partial matches OK)
-      *  <   less than
-      *  >   greater than
-      **/
 
     // Use regex to split string into a list, preserving quoted phrases as single value
     val matches = ListBuffer[String]()
-    val m = Pattern.compile("([^\"]\\S*|\".+?\")\\s*").matcher(query.replace(":", " "))
+    val m = Pattern.compile("([^\"]\\S*|\".+?\")\\s*").matcher(query)
```
lmarini, line 830: I noticed that white space between key, operator and value is not allowed. Could we add that so that the user can have extra white space such as:
"syn:loc_selector"  ==   "5310:123-456-7890"

lmarini:
Doing some more random testing:

"Fish:contactz"<"banchovy:223-456-7890" returns the file, "Fish:contactz">"banchovy:223-456-7890" does not.

max-zilla:
> Doing some more random testing:
> 
> "Fish:contactz"<"banchovy:223-456-7890" returns the file, "Fish:contactz">"banchovy:223-456-7890" does not.

I was playing with how Elasticsearch handles < > on string values and I wasn't getting the alphabetic behavior i was expecting. Will try to find specific documentation about it.

Not sure if I have control over including only documents with a given field, but might be a simple query parameter.

max-zilla:
> I am wondering if there is a way to make it so that "IMPORTANT NUMBER"!=34 only returns documents that have the field "IMPORTANT NUMBER" set but of which the value is not 34.

I believe an 'exists' clause could apply this, giving it a try to see if there are unexpected behaviors.

lmarini:
`"IMPORTANT NUMBER"<         35`

works and translates to 

`2020-08-27 16:55:55,963 - [INFO   ] - application - Searching Elasticsearch: {"bool":{"must":[{"range":{"metadata.IMPORTANT NUMBER":{"lt":"35"}}},{"bool":{"should":[{"match":{"child_of":"5eea7e53e4b0f613c815cfad"}},{"match":{"creator":"5eea791204410fb7bc3459a2"}}]}}]}}`

`"IMPORTANT NUMBER"               <35`

doesn't seem to work and translates to

```2020-08-27 16:54:48,243 - [INFO   ] - application - Searching Elasticsearch: {"bool":{"must":[{"regexp":{"_all":".*important number.*"}},{"range":{"_all":{"lt":"35"}}},{"bool":{"should":[{"match":{"child_of":"5eea7e53e4b0f613c815cfad"}},{"match":{"creator":"5eea791204410fb7bc3459a2"}}]}}]}}```

max-zilla:
> `"IMPORTANT NUMBER"< 35`
> 
> works and translates to
> 
> `2020-08-27 16:55:55,963 - [INFO ] - application - Searching Elasticsearch: {"bool":{"must":[{"range":{"metadata.IMPORTANT NUMBER":{"lt":"35"}}},{"bool":{"should":[{"match":{"child_of":"5eea7e53e4b0f613c815cfad"}},{"match":{"creator":"5eea791204410fb7bc3459a2"}}]}}]}}`
> 
> `"IMPORTANT NUMBER" <35`
> 
> doesn't seem to work and translates to
> 
> `2020-08-27 16:54:48,243 - [INFO ] - application - Searching Elasticsearch: {"bool":{"must":[{"regexp":{"_all":".*important number.*"}},{"range":{"_all":{"lt":"35"}}},{"bool":{"should":[{"match":{"child_of":"5eea7e53e4b0f613c815cfad"}},{"match":{"creator":"5eea791204410fb7bc3459a2"}}]}}]}}`

Good catch. There was an error in my regex so it was searching for `term ":` instead of `term" :` so it wouldn't parse this specific case directly. Adding similar queries to my test cases.

PR #66, opened 2020-09-02
Fix proxy and more

robkooper:
## Description

This fixes 3 things:
- if a proxy call ends with a / it will be stripped, it now leaves the / at the end
- if a path ends with a / it will now get rewritten with the correct prefix
- fix 2 missing , in the jsonld files

I tried to fix the code for untrail to catch the /api/proxy calls, but it became messy. Since the conf/routes is executed in order, moving these calls to the top of the file fixed the problem.

## Review Time Estimate

- [ ] Immediately
- [X] Within one week
- [ ] When possible

## Types of changes

## Checklist:

- [ ] I have signed the CLA
- [X] I have read the **CONTRIBUTING** document.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## Unreleased
+
+### Fixed
+- proxy will no correctly handle paths that end with a / at the end.
+
 ## 1.11.0 - 2020-08-31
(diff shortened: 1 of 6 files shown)


bodom0015 (approved):
LGTM :+1:

Unclear how bullet points 1 and 2 differ, but can confirm that the presence of a trailing slash no longer affects behavior

lmarini:
changelog missing

lmarini (approved):

lmarini:
Need to double check why I couldn't reproduce the issue with `application.context="/clowder/"` eating the context when redirecting with trailing slash on local.

PR #74, opened 2020-09-15
Overriding extractors at the space level when they are enabled at the global level.

lmarini:
## Description
Users now have a radio buttons at the space level to override an extractor enable at the global level. Code has changed behind the scenes so it is really important that we make sure it works as intended. Need help testing to make sure that the messages are going to the correct rabbitmq queue.

Not only can a user override a globally enabled extractor, but they can select to disable an extractor in the space even if it is not enabled globally (need a better term for this) in case it will be enabled at the global level in the future. Keep in mind that by default new spaces will have no selections set. This is due to how we store things in Mongo.

Screen Shot 2020-09-25 at 4 13 33 PM

To test this you have to drop `spaces.extractors` table in Mongo. You can then enable/disable spaces at the global level and override them at the space level.

TODO:
- Need to add update script to update Mongo schema.
- If possible I would like to order the list of extractors so that the ones enabled at the global level always show up at the top

## Review Time Estimate

- [ ] Immediately
- [x] Within one week
- [ ] When possible

## Types of changes

- [X] New feature (non-breaking change which adds functionality)

## Checklist:

- [ ] I have signed the CLA


lmarini:
Added mongo update script.

bodom0015 (commented):
I'm running into trouble testing this one. I have 2 spaces, titled "Disable All the Things" and "Stuff is Enabled Here", with my extractor enabled globally. Despite disabling my extractors in the so-named space, uploading a file to a dataset in that space still triggers the extractor.

Stepping through, it looks like `getQueuesFromBindings` might be re-enabling the extractor? The comment suggests that this method is meant to support legacy behavior, but even deleting my existing queues and allowing clowder to recreate them didn't appear to change the behavior.

See https://recordit.co/Xlhfyj4JoX

Have you seen this as well? Any idea why this legacy method might be triggering on a new extractor, or have I perhaps forgotten some aspect in clearing out the old state from RabbitMQ?

lmarini:
> I'm running into trouble testing this one. I have 2 spaces, titled "Disable All the Things" and "Stuff is Enabled Here", with my extractor enabled globally. Despite disabling my extractors in the so-named space, uploading a file to a dataset in that space still triggers the extractor.
> 
> Stepping through, it looks like `getQueuesFromBindings` might be re-enabling the extractor? The comment suggests that this method is meant to support legacy behavior, but even deleting my existing queues and allowing clowder to recreate them didn't appear to change the behavior.
> 
> See https://recordit.co/Xlhfyj4JoX
> 
> Have you seen this as well? Any idea why this legacy method might be triggering on a new extractor, or have I perhaps forgotten some aspect in clearing out the old state from RabbitMQ?

@bodom0015 thank you for bringing this up. It took a while for @robkooper and I to figure this out, but I think we know whats going on.

The simple answer is, the extractors currently add bindings (the old way that you pointed out is still there for backward compatibility). I think one way to test this is pull up the rabbitmq management console and remove the bindings for the queues. getQueuesFromBindings should then be empty (you will need to restart Clowder, they are loaded at startup).

The long answer is that we will change the default behavior of pyclowder to create the binding in a dummy exchange. We will also fix a bug in Clowder that is making the bindings not shop up when running in the stack in docker compose. This is why I didn't see the error in my tests. Are you testing with Rabbitmq running on your host and not in a container?

Hopefully we can stop supporting this backward compatibility soon(tm).

bodom0015 (changes requested):
@lmarini I have Mongo/RabbitMQ/Elasticsearch running in a raw Docker container (not using compose), with Clowder running in IntelliJ

I think that if there are additional bugs that need to be fixed for this behavior to work as expected, then those issues should be addressed before merging this up to develop.

lmarini:
> @lmarini I have Mongo/RabbitMQ/Elasticsearch running in a raw Docker container (not using compose), with Clowder running in IntelliJ

Does your rabbitmq management console run under http://localhost:15672/rabbitmq or http://localhost:15672?

> I think that if there are additional bugs that need to be fixed for this behavior to work as expected, then those issues should be addressed before merging this up to develop.

No bugs are introduced with this PR. The issue is on master (clowder/pyclowder).

max-zilla:
I had same issue Mike did until I stopped Clowder, removed all bindings for the extractor, ran the extractor in docker, and restarted Clowder.

After that, enabling wordcount but disabling it on space level meant any text files i uploaded to that space didn't get submitted but the normal default behavior space did get them.

max-zilla:
Changelog is empty, otherwise I am approving.

max-zilla (approved):
Tested as described in my comment and it looks good.

Don't forget CHANGELOG.

PR #76, opened 2020-09-16
Add numeric and date fields, gt lt gte lte operators

max-zilla:
!!This PR will require a reindex!! POST to /api/reindex after adding some metadata that are numbers or something like 2011-12-12 and the range searches should handle them as expected.

It enables numeric and date detection for new metadata fields and adds < > <= >= operators for that comparison. 

Those operators had been disabled when string parsing was the only enabled detection, because that would consider "10" to be less than "9" given the alphabetical order instead of numerical.


lmarini:
Missing changelog entry and message about elasticsearch reindex required.

bodom0015 (commented):
Can confirm that numerical values work as described, but I am unable to add any "Date and Time" metadata values. It appears the datepicker is never initialized, and the field is just a text input? Entering a value manually doesn't appear to work either, as the resulting term appears to be blank.

Is anyone else seeing this behavior as well? I don't really see anything in this PR that could have caused this... Should I file this as a new bug/issue?

My test case:
1. Add 2 new metadata terms representing a String and a Datetime, call them `number` and `date` respectively
2. Attach metadata to a file:
  * `number: 6`
  * ~~`date: 01/20/2020`~~ I was unable to add a value here
3. Navigate to Metadata Search
4. Search for files with `number` greater than `10` - this should return no results
5. Search for files with `number` less than `10` - this should return at least 1 result, including the file we just attached it to

max-zilla:
@bodom0015 thanks for responses. for all of these, make sure you've done a reindex first. if you go to elasticsearch mappings (http://localhost:9200/clowder/_mappings if youre running locally) you should see `numeric_detection: true,` near the top and nothing about date_detection (which is true by default).

1 - no datepicker implemented yet. you can use search string in date format, but Elasticsearch has some specific date formats it looks for: https://www.elastic.co/guide/en/elasticsearch/reference/current/date.html
I uploaded following metadata:
Screen Shot 2020-10-02 at 8 11 08 AM
...first and third were recognized as dates in _mapping but middle one (your example) was not, i think this is consistent with their expected format:
Screen Shot 2020-10-02 at 8 11 31 AM

...however, deeper testing with different formats shows that ES validates query string against the date format specified, which means if the field is YYYY/MM/DD and you query YYYY-MM-DD you get an error, and vice versa:
`failed to parse date field [2020-01-29] with format [yyyy/MM/dd HH:mm:ss||yyyy/MM/dd||epoch_millis]`

So if users mix and match their formatting this is annoying:
Screen Shot 2020-10-02 at 8 23 57 AM

This probably merits another week to add a fix. Two that come to mind:
A - when a date field is specified in query, before sending to ES we parse the date in scala and change it into matching format for the field. this would help both GUI and API searches.
B - add a proper datepicker to GUI that appears when you select a date field from autocomplete instead of a text box. But this doesn't help API users or the search box in upper right.

We can discuss more at this week's meeting.

max-zilla:
@lmarini i updated the help text for this last week, prior to that everything was working during our testing, can we merge this today if possible?

PR #78, opened 2020-09-17
run in dedicated context

robkooper:
- can specify the parallelism
- can specify batch size

this also adds an index that might speed things up

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [X] When possible

## Types of changes

## Checklist:

- [ ] I have signed the CLA


app/services/mongodb/MongoDBQueueService.scala
```diff
@@ -68,38 +69,36 @@ trait MongoDBQueueService {
...
   def listen() = {
-    if (queueTimer == null) {
-      // TODO: Need to make these in a separate pool
-      queueTimer = Akka.system().scheduler.schedule(0 seconds, 5 millis) {
-        getNextQueuedAction match {
-          case Some(qa) => handleQueuedAction(qa)
-          case None => {}
-        }
+    if (ec == null) {
+      ec = ExecutionContext.fromExecutor(new java.util.concurrent.ForkJoinPool(threads))
```
lmarini, line 97: What's the difference between this and creating and Akka context with `implicit val myExecutionContext: ExecutionContext = Akka.system.dispatchers.lookup("my-context")
`?

I looked it up a bit but couldn't find a clear answer. Thank you.

lmarini: This is also an interesting read https://jessitron.com/2014/01/29/choosing-an-executorservice/.

robkooper: only issue I saw/had was when using the loopup you need to define it in your app.config, otherwise it throws an error. in this case it creates one for N threads, and runs the indexer in parallel. This was hoping to be able to process a large number of waiting messages quickly using multiple threads.

robkooper:
@max-zilla and @lmarini I want to discuss this in more detail. Marking as draft for now and moving the index creation in a new PR.

robkooper:
closing this PR for now, but keeping branch.

The thinking was to have multiple threads processing the queue to speed up a reindex. With the changes to elasticsearch not sure if that is wantable. 

There is another issue that is more critical and might be some of the issues we seeing with speed. If we have multiple instances of clowder, each of them could potentially be trying to reindex the same object.

PR #80, opened 2020-09-25
Extractor Catalog - enable trigger filter to separate file/dataset extractors

max-zilla:
This adds a small bit of logic to the /extractors endpoint that will filter by extractor trigger if file/* or dataset/* is selected. In full implementation this should have dynamic autocomplete or at least a list generated from found records, but as a proof-of-concept here is minimum example.


lmarini:
Needs changelog.

lmarini (changes requested):
I don't know if this needs to be address in this PR, but if I select a filter for which I have no extractors matching, the page will show blank with a "No extractors running" message. This was ok before when the list of extractors represented that, but it's confusing now that we filter down. Should we show the regular page with no extractors in the table?

max-zilla:
> I don't know if this needs to be address in this PR, but if I select a filter for which I have no extractors matching, the page will show blank with a "No extractors running" message. This was ok before when the list of extractors represented that, but it's confusing now that we filter down. Should we show the regular page with no extractors in the table?

I noticed that with metadata filter as well. Probably better to show the regular page with no extractors once a filter is chosen, will see if i can modify.

max-zilla:
> Needs changelog.

this is into Mike's branch/PR, not develop. i think that changelog will cover everything

bodom0015 (approved):
LGTM, aside from hitting the same "No extractors running" issue as mentioned above.

Works as advertised :+1:

max-zilla:
OK, I simply removed the "No extractors found" page and we'll always show the list. I re-added the metadata filter to demonstrate this as many instances will not have any, but it isn't as disruptive now. If the instance truly has no extractors running, they'll see an empty list with filter controls. (commit looks larger than it is due to it being mostly a large de-dent of the indent level)

Issue #84, opened 2020-09-27
All clowder tests are failing for BrownDog since v1.11.0


robkooper:
**Describe the bug**

The BrownDog tests are failing since version 1.11.0

Upload would succeed, but extraction would return:
```
Exception in extracting metadata code=500 msg='{"status":"KO","msg":"No jobId found for Extraction on fileid=5f70e2624f0c5995f50f78cb"}'
```

PR #87, opened 2020-09-29
Release/1.11.1

lmarini:
### Added
- Added healtz endpoint that is cheap and quick to return, useful for kubernetes live/ready checks.

### Fixed
- Fixed health check script when using custom path prefix.
- Proxy will no correctly handle paths that end with a / at the end.
- Submitting an extraction will always return a 500 error, see [#84](https://github.com/clowder-framework/clowder/issues/84)
- Added MongoDB index for `folders.files`.

### Changed
- Updated update-clowder script to work with migration to github. Has the ability now to push a message to MSTEAMS as well as influxdb.

## Review Time Estimate

- [x] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:

- [ ] I have signed the CLA

### DIFF
--- a/docker/healthcheck.sh
+++ b/docker/healthcheck.sh
@@ -1,3 +1,6 @@
 #!/bin/bash
 
-curl -s --fail http://localhost:9000/api/status || exit 1
+### Add trailing backslash if not in context
+[[ "${CLOWDER_CONTEXT}" != */ ]] && CLOWDER_CONTEXT="${CLOWDER_CONTEXT}/"
+
+curl -s --fail http://localhost:8000${CLOWDER_CONTEXT:-/}api/status || exit 1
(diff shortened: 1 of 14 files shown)


docker/healthcheck.sh
```diff
@@ -1,3 +1,6 @@
 #!/bin/bash
 
-curl -s --fail http://localhost:9000/api/status || exit 1
+### Add trailing backslash if not in context
+[[ "${CLOWDER_CONTEXT}" != */ ]] && CLOWDER_CONTEXT="${CLOWDER_CONTEXT}/"
+
+curl -s --fail http://localhost:8000${CLOWDER_CONTEXT:-/}api/status || exit 1
```
robkooper, line 6: should use the new healtz endpoint it is cheaper to call.

robkooper (approved):

docker/healthcheck.sh
robkooper: See #88

PR #69, opened 2020-09-08
Extractor Catalog: Extractor Maturity features

bodom0015:
### Problem
We need a way to mark an extractor as "Development" / "Staging" / "Production" to signal to users whether the extractor is ready for public consumption

### Approach
* Add `maturity` field to ExtractorInfo model
* ~~Add an API method to change the maturity: `PUT /extractors/{name}/maturity/{maturity}`~~
* Display the `maturity` field value in the Maturity column on the Extractor Catalog
* Add helper functions to `extractors.js` for fetching / updating ExtractorInfo via API
* Wire up dropdown buttons for "Move to Development" / "Move to Production" / etc to the API

### How to Test
Prerequisites: at least one extractor registered, RabbitMQ plugin enabled, at least one admin user

1. Checkout and run this branch locally
2. Login to Clowder as an admin user
3. From the top navbar, choose Explore > Extractors
    * You should be brought to the Extractor Catalog view
    * You should see that each extractor has a Maturity value in the table
    * Maturity for all extractors should default to "Development"
4. Click the three vertical dots to the right of the extractor in the table
    * You should see the More Actions dropdown menu expand
    * You should see options to "Move to Staging" and "Move to Production"
5. Click "Move to Staging" to update the maturity
    * You should see the view refresh
    * You should see the Maturity value in the table has now updated
6. Expand the dropdown again
    * You should see that "Move to Development" now appears
    * You should see that "Move to Staging" no longer appears in the list

## Review Time Estimate

- [ ] Immediately
- [x] Within one week
- [ ] When possible

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue) 
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [ ] My change requires a change to the documentation.
- [ ] I have updated the CHANGELOG.md.
- [ ] I have signed the CLA
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #90, opened 2020-09-30
Feature/extractor info update date

### LINKED ISSUE #89: `POST /api/extractors` should ignore `update` field
The field should we be set internally so that a user cannot overwrite the update date of an extractor info document with an arbitrary date time.

Also when printing out and ExtractorInfo document the date fields should be printed using a standard format.

lmarini:
## Description
When a client currently does a `POST /api/extractors` it can provide a value for the `update` field. This shouldn't be allowed as this is an internal date to keep track of when the document was last updated.

This fix makes it so that if `update` is set, it will be ignored and it will be automatically be set to the current date time.
This also prints the update date when doing a `GET /api/extractors` using ISO08601.

Fixes #89. 

## Review Time Estimate

- [ ] Immediately
- [x] Within one week
- [ ] When possible

## Types of changes

## Checklist:

- [x] I have signed the CLA


CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Script to clean extractors' tmp files.
 - Script for RabbitMQ error queue cleanup.
 - Ability to use basic html formatting in the welcome message on the home page. [#51](https://github.com/clowder-framework/clowder/issues/51)
+- Ignore the `update` field when posting to `/api/extractors`. [#89](https://github.com/clowder-framework/clowder/issues/89)
```
robkooper, line 14: looks like the changelog is old

app/models/Extraction.scala
```diff
@@ -121,16 +124,21 @@ object ExtractorInfo {
     def reads(json: JsValue): JsResult[URL] = JsSuccess(new URL(json.toString()))
     def writes(url: URL): JsValue = Json.toJson(url.toExternalForm)
   }
+
   implicit val dateFormat = new Format[Date] {
+    val simpleDateFormat = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSS'Z'")
```
robkooper, line 129: doesn't look like we use simplDateFormat

robkooper (changes requested):

CHANGELOG.md
lmarini: it's going into `extractor-catalog-wireframe`. I just merged that branch in this one.

app/models/Extraction.scala
lmarini: Thanks. I switched to using the function in `util.Parsers` at the last minute. Will fix.

bodom0015 (approved):
LGTM :+1: Works as advertised, and resolves the issue uncovered in #69

app/models/Extraction.scala
lmarini: Removed.

PR #98, opened 2020-10-09
Enable ingest of a Bagit zipfile

max-zilla:
This enables user to upload zip file to new endpoint to create a Clowder dataset with the contents:
`curl -XPOST -F file=@MyBag.zip http://localhost:9000/api/datasets/createfrombag?key=foo`

I've been testing with Clowder bags and it will bring along metadata, still need bugfixes and testing with external bagit next week.


max-zilla:
I have some Dataverse-exported bags I can share with anyone testing this, you can also download a bag (dataset) from a Clowder instance and test with that too.

lmarini (approved):
Works great. Removed a few files that were accidentally checked in and added an issue to make the endpoint return JSON instead of plain text.

https://github.com/clowder-framework/clowder/issues/208

Issue #83, opened 2020-09-25
Make search public


lmarini:
Currently the search page is only available to users who are logged in. Now that we have the ability of keep some data private and some data public, the search page should be available to users who are not logged in. Same with the API endpoint.

We need to double check permissions of what gets returned and make sure making the page public doesn't create any issues.

MBcode:
http://localhost:9000/metadata/search has been available w/o logging in; so assume search fix closed this

PR #100, opened 2020-10-12
Fix several search result bugs

max-zilla:
This addresses several issues with /api/search and /search pages:

- SuperAdmin mode was not respected by the metadata search/search API/search box. Now enabling superAdmin will show all results in DB that match.
- The search omission message wasn't behaving properly at the end of the autoscroll. It would erroneously say it had omitted results on the final page even if it had not (i.e. with superAdmin fix above) due to an edge case bug. I fixed the bug in the code, but...
- due to the change in permissions handling on service, the user will NO LONGER SEE OMITTED COUNTS. i.e. if you search "test" without permissions there might be 111 results in Db for superadmin but you'll see 0. we can discuss if this is controversial. fix applies to both search GUI pages.
- fixed a dev line that snuck in from a previous PR that was restricting the search GUI to fetching results 2 at a time (!). this didn't affect API calls.


app/api/Search.scala
```diff
@@ -1,7 +1,7 @@
 package api
 
 import api.Permission._
-import services.{RdfSPARQLService, DatasetService, FileService, CollectionService, PreviewService, SpaceService,
+import services.{RdfSPARQLService, PreviewService, SpaceService,
```
robkooper, line 4: probably cleaner to combine line 4 and 5.

CHANGELOG.md
```diff
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## Unreleased
+
+### Fixed
+- Fixed permissions checks on search results for search interfaces that would cause misleading counts.
```
robkooper, line 10: can you add a link to the superadmin issue here.

app/services/ElasticsearchPlugin.scala
```diff
@@ -807,6 +808,28 @@ class ElasticsearchPlugin(application: Application) extends Plugin {
         builder.startObject().startObject("exists").field("field", key).endObject().endObject()
       })
 
+      // If user is superadmin or there is no user, no filters applied
```
robkooper, line 811: this comment seems strange, this makes it sound like if a user is not logged in they will get same list back as superadmin.

app/views/searchResults.scala.html
```diff
@@ -108,7 +108,6 @@ <h1>Search</h1>
     <script>
         // count_stepsize is the batch size to get results; e.g. 60 would be 1-60, 61-120, 121-180, etc.
         var count_stepsize = @{play.api.Play.configuration.getInt("elasticsearchSettings.maxResults").getOrElse(240)};
-        count_stepsize = 2;
```
robkooper, line 111: 💯

robkooper (commented):
need to run code next.

app/services/ElasticsearchPlugin.scala
max-zilla: At the moment all the endpoints for search require login, so right now user would have no path to this code. If in future someone made /search public, then without a user yes we are missing a piece in elasticsearch to filter by public. 

we could potentially get a list of spaces that are public and apply those as the filter for that case, but since the code can't currently be reached i did not do so.

max-zilla: ok, will work on adding the public spaces filter for people not logged in.

max-zilla: I added some basic handling in there but will need to be verified during testing #83.

bodom0015 (approved):
Appears to work as advertised :+1:

lmarini (changes requested):
Not sure if this is working as intended but if I do a search as a user who doesn't have access to much data, I will still get a number of results, even if I don't have access to them. For example:

Screen Shot 2020-10-28 at 4 45 56 PM

If I go super admin, I see a different count and all the files:

Screen Shot 2020-10-28 at 4 48 06 PM

Any thoughts of what I can check?

max-zilla:
> Any thoughts of what I can check?

Hmm, that is not working as intended, I expect to see 0 in the first case (or, 8 with them actually listed). This could mean elasticsearch finds resources that pass its ownership filter:

```// Restrict to spaces the user is permitted access to
            permitted.foreach(ps => {
              builder.startObject().startObject("match").field("child_of", ps.stringify).endObject().endObject()
            })

            // Also include anything the user owns
            builder.startObject().startObject("match").field("creator", u.id.stringify).endObject().endObject()```

...but those resources are subsequently rejected by Clowder's MongoDB permission check:

```val filesList = if (superAdmin)
      files.get(response.results.filter(_.resourceType == 'file).map(_.id)).found
    else
      files.get(Permission.checkPermissions(user, Permission.ViewFile,
        response.results.filter(_.resourceType == 'file)).approved.map(_.id)).found```

This PR shouldn't require a reindex, although since you did one that counts out a few possibilities (elasticsearch having records for files that no longer exist, for example). Two things to check:

- What do you see with /api/search e.g. http://localhost:9000/api/search?query=test ? 
- in api.Search search() around line 58 if you add a log:
```val response = plugin.search(query, resource_type, datasetid, collectionid, spaceid, folderid, field, tag, from_index, size, permitted, request.user)
        Logger.debug(response.results.toString) // ADD THIS
        val result = SearchUtils.prepareSearchResponse(response, source_url, request.user)
        Ok(toJson(result))```

...you should see the 8 IDs elasticsearch finds that you aren't seeing. can you view these in Clowder? anything unusual?

I have no meetings today (Thursday) so we can hop on a call to debug for 20 mins if you want too.

max-zilla:
Had a call with @lmarini and we tracked this down to a discrepancy with space permissions vs. file permissions. Created an issue here: https://github.com/clowder-framework/clowder/issues/125 and removed total result count from search results, this  should avoid user confusion until better solution is possible.

lmarini (approved):

lmarini (approved):

Issue #94, opened 2020-10-05
Log user who submitted an extraction


lmarini:
We currently don't track the user in the extractor event.

Issue #60, opened 2020-09-02
search as superadmin does not show all results


robkooper:
**Describe the bug**
When performing a search as super admin it will indicate that N items are found, but filters them based on your access rights not based on if you have super admin enabled. 

**Expected behavior**
The expectation is that if I'm super admin, search should show everything that I can see.

max-zilla:
https://github.com/clowder-framework/clowder/pull/100 there is a PR open now that fixes this and a handful of other issues with search.

PR #79, opened 2020-09-18
extractor view metrics

bingzhang:
## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [x] When possible

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue) 
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [ ] My change requires a change to the documentation.
- [x] I have updated the CHANGELOG.md.
- [ ] I have signed the CLA
- [ ] I have updated the documentation accordingly.
- [ ] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #108, opened 2020-10-14
Release/1.12.0

lmarini:
## 1.12.0 - 2020-10-19
**_Warning:_**
- This update modifies the MongoDB schema. Make sure to start the application with `-DMONGOUPDATE=1`.
- This update modifies information stored in Elasticsearch used for text based searching. Make sure to initiate a reindex 
  of Elasticsearch from the Admin menu or by `POST /api/reindex`.

### Added
- Global extractors page now shows more information, including submission metrics, logs (using Graylog), job history and
  extractors maturity. Extractors can be grouped using labels. User can filter list of extractors by labels, space, trigger
  and metadata key.
- Users have more refined options to set extractors triggers at the space level. They can now follow global settings, 
  disable and enable triggers.
- Ability to set chunksize when downloading files. Set defult to 1MB from 8KB. This will result in faster downloads and 
  less CPU usage at the cost of slightly more memory use.
- Support for parsing of Date and Numeric data in new metadata fields. New search operators <, >, <=, >= have been 
  added to search API now that they can be compared properly.
- Track user_id with every extraction event. [#94](https://github.com/clowder-framework/clowder/issues/94)
- Added a new storage report at `GET api/reports/storage/spaces/:id` for auditing user storage usage on a space basis.
- The file and dataset metrics reports also have support for since and until ISO8601 date parameters.

### Fixed
- Ignore the `update` field when posting to `/api/extractors`. [#89](https://github.com/clowder-framework/clowder/issues/89)
- Search results were hardcoded to be in batches of 2.
- Fixed permissions checks on search results for search interfaces that would cause misleading counts. [#60](https://github.com/clowder-framework/clowder/issues/60)


max-zilla:
Pushed fix for the stepsize bug.

bodom0015:
@lmarini for extractors running the newer base image / version of pyclowder, this should reflect the number of unique Jobs submitted:
ExtractorCatalogJobCount

ExtractorCatalogJobCountsDB

For older extractors that do not assign a `job_id` to extractions, this would only count the total number of messages

max-zilla:
Tested extractor pages, range searches, extractor space override. didn't encounter any issues after an hour or so, approving.

max-zilla (approved):

app/services/mongodb/MongoDBExtractionService.scala
```diff
@@ -39,6 +41,12 @@ class MongoDBExtractionService extends ExtractionService {
     Extraction.find(MongoDBObject("file_id" -> new ObjectId(resource.id.stringify))).toList
   }
 
+  def findByExtractorIDBefore(extractorID: String, status: String, date: String, limit: Int): List[Extraction] = {
+    val order = MongoDBObject("start"-> -1 )
+    val sinceDate = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSS").parse(date)
+    Extraction.find($and("extractor_id" $eq extractorID, "status" $eq status, "start" $gte sinceDate)).sort(order).limit(limit).toList
+  }
```
bodom0015, line 48: This still appears to be counting the number of events, not the number of unique jobs. I am not sure how to run [`distinct` queries](https://docs.mongodb.com/manual/reference/method/db.collection.distinct/) with Casbah, but am looking into it now. 

This was pointed out during the review, but never addressed: https://github.com/clowder-framework/clowder/pull/79/files#r499039736

bodom0015 (changes requested):

app/services/mongodb/MongoDBExtractionService.scala
lmarini: You can have access to the underlying casbah collection through the DAO `Extraction.dao.collection.distinct()`.

app/controllers/Extractors.scala
```diff
@@ -80,6 +112,89 @@ class Extractors  @Inject() (extractions: ExtractionService,
...
+    if(n > 0) {
+      average = average/n
+    }
+    Logger.debug("average: " + average)
+    // get at the number of ``SUBMITTED'' submission in last week.
+    val lastweeksubmitted = extractions.findByExtractorIDBefore(extractorName, "SUBMITTED", last7daydate, 0)
+    Logger.debug("lastweek submitted: " + lastweeksubmitted.size)
+    // get at the number of ``SUBMITTED'' submission in last month.
+    val lastmonthsubmitted = extractions.findByExtractorIDBefore(extractorName, "SUBMITTED", lastmonthdate, 0)
+
```
bodom0015, line 166: This should count the number of unique jobs submitted, instead of the number of raw events we've seen:
```suggestion
    Logger.debug("average: " + average)
    // get at the number of ``SUBMITTED'' submission in last week.
    val lastweekevents = extractions.findByExtractorIDBefore(extractorName, "SUBMITTED", last7daydate, 0)
    val lastweeksubmitted = lastweekevents.map(evt => evt.job_id).distinct.length
    Logger.debug("lastweek submitted: " + lastweeksubmitted)
    // get at the number of ``SUBMITTED'' submission in last month.
    val lastmonthevents = extractions.findByExtractorIDBefore(extractorName, "SUBMITTED", lastmonthdate, 0)
    val lastmonthsubmitted = lastmonthevents.map(evt => evt.job_id).distinct.length

```

app/controllers/Extractors.scala
```diff
@@ -80,6 +112,89 @@ class Extractors  @Inject() (extractions: ExtractionService,
...
+    var lastTenAverage = BigDecimal.valueOf(sum)
+    Logger.debug("last 10 average: " + lastTenAverage)
+    if(n > 0) {
+      lastTenAverage = lastTenAverage/n
+    }
+    Logger.warn("last 10 average: " + lastTenAverage)
+
+    val targetExtractor = extractorService.listExtractorsInfo(List.empty).find(p => p.name == extractorName)
+    targetExtractor match {
+      case Some(extractor) => Ok(views.html.extractorMetrics(extractorName, average.toString, lastTenAverage.toString, lastweeksubmitted.size, lastmonthsubmitted.size))
```
bodom0015, line 193: This should count the number of unique jobs submitted, instead of the number of raw events we've seen:
```suggestion
      case Some(extractor) => Ok(views.html.extractorMetrics(extractorName, average.toString, lastTenAverage.toString, lastweeksubmitted, lastmonthsubmitted))
```

bodom0015 (changes requested):
Thank you, @lmarini. That was exactly what I was looking for.

Unfortunately, `Extraction.dao.collection.distinct("job_id", query)` appears to return a List of only the unique job_ids. Since we need the actual `Extraction` object's `start` Date to calculate the average runtime, this may not be the right approach here.

We can use Scala's built-in `map` and `distinct` to count the number of unique entries. This should not affect the claculation of the runtimes, as the `lastweeksubmitted` and `lastmonthsubmitted` are not used in this calculation loop.

CHANGELOG.md
```diff
@@ -4,10 +4,32 @@ All notable changes to this project will be documented in this file.
...
+  faster downloads and less CPU usage at the cost of slightly more memory use.
+- Support for parsing of Date and Numeric data in new metadata fields. New search operators <, >, <=, >= have been 
+  added to search API now that those data are properly compared.
+- Global extractors page now shows more information, including submission metrics, logs (using Graylog), job history and
+  extractors maturity.
+- Users have more refined options to set extractors triggers at the space level. They can now follow global settings, 
+  disable and enable triggers.
+
+### Fixed
+- Remove secretKey from extractor log.
```
robkooper, line 24: not sure we need to mention that here since it is a fix to something that was never released.

bodom0015 (approved):

max-zilla:
I approved this 2 weeks ago, but just tested it more with the new code and verified the extractor search/labels and the new reporting search are all as expected so I came here to re-approve.

PR #114, opened 2020-10-21
Space-based file storage report endpoint

max-zilla:
/api/reports/storage/spaces/:id endpoint with file records for user-based storage tracking

Screen Shot 2020-10-21 at 8 56 30 AM

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Add endpoint and view page for the extractor metrics.
 - Add endpoint and view page to view a extractor's log from graylog.
 - Ignore the `update` field when posting to `/api/extractors`. [#89](https://github.com/clowder-framework/clowder/issues/89)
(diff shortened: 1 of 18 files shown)


lmarini (approved):
Looks good. Two minor comments:

- The column `space_owners` shows `0` for all resources and the column `space_admins` is empty.
- I don't see updates to the swagger documentation

max-zilla:
@lmarini I updated Swagger and fixed a bug where space and role was stored capitalized "Admin" instead of "admin". However I could not replicate space owners as 0, do you actually have an owner ID for the space in question in mongo? It should at minimum include the space.creator ID for the space you requested in the report.

lmarini:
> @lmarini I updated Swagger and fixed a bug where space and role was stored capitalized "Admin" instead of "admin". However I could not replicate space owners as 0, do you actually have an owner ID for the space in question in mongo? It should at minimum include the space.creator ID for the space you requested in the report.

Thank you. Both are fixed. One was due to a bad document in mongodb.

PR #120, opened 2020-10-27
Extractor Catalog: search filters + matching ANY/ALL

bodom0015:
### Problem
Design includes more ways to filter catalog results than we currently offer.

### Approach
* Add generic search filter to the Extractor Catalog
* Add space search filter to the Extractor Catalog
* Add metadata search filter to the Extractor Catalog
    * NOTE: this assumes `ExtractorInfo.contexts` has been populated properly.. for example, the wordcount extractor contains the proper entries
* Option to match either ALL filters or ANY filter
* Make sure these new filters play nicely with the label filter

### How to Test
Prerequisites: RabbitMQ plugin enabled, at least two different types of extractors registered, wordcount extractor registered

1. Checkout and run this branch locally
2. Login to Clowder
3. Navigate to the Extractor Catalog
    * You should see a text input allowing for a general text search (e.g. name, authors, description, etc)
    * You should see radio buttons offering to "match ANY" or "match ALL"
4. Enter a nonsensical combination, for example: `Match ALL` and a general text search of `asdf`
    * You should see no results
5. Choose a label (e.g. Image), `Match ALL`, and general search of "Rob Kooper"
    * You should see only image-based extractors that include "Rob Kooper" as a contributor or author
6. Choose no label, enter "Image" as generic search with `Match ANY` and choose "words" from the metadata dropdown
    * You should be shown any extractors mentioning "Image", as well as the wordcount extractor
7. Same as 6, except choose `Match ALL` this time
    * You should see no results

NOTE: If a Label is selected, then we calculate its intersect with the results of the other filters regardless of whether "match ANY" / "match ALL" is selected. This is to avoid the case where the user chooses "Image" explicitly, where "match ANY" will can non-image results to that list.

## Review Time Estimate

- [ ] Immediately
- [x] Within one week
- [ ] When possible

## Types of changes

## Checklist:

- [ ] I have signed the CLA
- [x] All new and existing tests passed.

### DIFF
--- a/app/controllers/Extractors.scala
+++ b/app/controllers/Extractors.scala
@@ -64,6 +64,12 @@ class Extractors  @Inject() (extractions: ExtractionService,
     val allLabels = extractorService.listExtractorsLabels()
     val categorizedLabels = allLabels.groupBy(_.category.getOrElse("Other"))
 
+    // Default value for unmatched filter
+    val defaultValue: List[ExtractorInfo] = request.getQueryString("matching") match {
+      case Some("all") => runningExtractors
+      case _ => List[ExtractorInfo]()
+    }
+
     // TODO: Autocomplete on text field from dynamically generated list of distinct triggers
     val triggerMatches: List[ExtractorInfo] = request.getQueryString("processTriggerSearchFilter") match {
       case Some("file/*") => runningExtractors.filter(re => re.process.file.length > 0)
@@ -72,7 +78,7 @@ class Extractors  @Inject() (extractions: ExtractionService,
       /*case Some(filt) => {
         // TODO: Need to figure out how to effectively search the ProcessTriggers structure
       }*/
-      case _ => runningExtractors
+      case _ => defaultValue
     }
 
     // Stringify full resource to perform simple search for user's query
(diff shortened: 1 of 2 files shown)


lmarini (changes requested):
I played around with different options and it looks like everything is working as intended. I really like that you are using the metadata definitions in the context for populating the metadata list. It's not something that we are good at keeping documented, but this is a great use for it.

The only thing that I would change is default listing. Right now it seems like it doesn't show anything by default when you open up the page. I would expect it to show all of the extractors and then the user could filter down the list using the controls. I am not sure a user would know what to filter on the first time they load this page. Similar to what we do for labels.

bodom0015:
Good catch! I made a commit to address this case.

The ALL + no filters case works because it starts with the full list and trims it down. ANY + no filters does not work in the same way because we start with an empty list and build it up.

More details:
The set theory works out well for the `ALL` case when no filters are entered. For ANY, things get a little weird if you don't specify one or more of the filters: blindly returning the full list in an ANY query would always return the full list, so we return an empty list for that case's default instead. This is why the `defaultValue` stuff was needed in this PR: intersect needs to be inclusive and union needs to be exclusive.

max-zilla:
Everything worked well for me! Also changing this to be against the Release branch.

max-zilla (approved):

lmarini (approved):

PR #124, opened 2020-10-28
initial implementation of extractor usage report

max-zilla:
Adds a new /api/reports/metrics/extractors endpoint with since and until flags, e.g.

http://localhost:9000/api/reports/metrics/extractions?since=2020-08-12T18:05:45.582Z

Due to database design (we don't store user data in the extractions table) we have to get those data from the datasets to make that association and the query is likely to be slow and taxing - we should discuss.


max-zilla:
To test this, unless you have a very new database from last month or so, you need to run update script to add user_id to your extractions:
`mongo clowder --quiet scripts/updates/UpdateUserId.js`

Then you can hit the api:
`http://localhost:9000/api/reports/metrics/extractors?since=2020-11-01T18:05:45.582Z`

lmarini (approved):

PR #126, opened 2020-11-02
Metadata tab select

MBcode:
## Review Time Estimate

- [x] Immediately
- [ ] Within one week
- [ ] When possible

## Types of changes

## Checklist:

- [ ] I have signed the CLA


MBcode:
When there are no files on dataset page, select metadata tab instead.

MBcode:
I was looking through the dataset templating file(app/views/dataset.scala.html line256)& noticed:  
@* FIXME Commented out since it creates some issues with dataset previewers *@
  @<script src=“@routes.Assets.at(“javascripts/tabsInURL.js”)” type=“text/javascript”></script>@
So (maybe) we should coordinate on the hash part of this (later)?

MBcode:
tabsInURL.js will take/set hash's from/to the URL. logic is working for setting active, but this js won't do the set-active part alone

MBcode:
the hash-url fixme warning probably should be better tested, but still want to get in the tab switching to metadata (when there, &no files).  Nice to get in soon, as external search is now directing these dataset pages, to see the metadata('details')

MBcode:
Not easy to test the hash-url part w/geoserver right now; So anyone is welcome to try out the split-out tab-switching (I'll make this active for review)

app/views/dataset.scala.html
```diff
@@ -269,11 +269,16 @@ <h4 class="modal-title" id="folder-modal-label">  @Messages("create.header", "Fo
 
 	    	<div class="tabbable" id="bottomDatasetTabbable"> <!-- Only required for left/right tabs -->
                 <ul class="nav nav-tabs margin-bottom-20" >
-                    <li role="presentation" class="active"><a href="#tab-files" role="tab" data-toggle="tab">Files </a></li>
-                    <li role="presentation"><a href="#tab-metadata" role="tab" data-toggle="tab">Metadata </a></li>
-                    <li role="presentation"><a href="#tab-extractions" role="tab" data-toggle="tab">Extractions </a></li>
-                    <li role="presentation"><a href="#tab-visua" role="tab" data-toggle="tab">Visualizations</a></li>
-                    <li role="presentation"><a href="#tab-comments" role="tab" data-toggle="tab">Comments (@comments.size)</a></li>
+										@if(dataset.files.isEmpty && (m.size > 0)) {
```
lmarini, line 272: Indentation is off. Thanks!

lmarini (changes requested):
We need to test if the dataset geospatial layers is affected by this. It wasn't the reason why this was turned off, but it is another previewer where the issue can happen.

lmarini (changes requested):
When pulling up the page with no files and some metadata, the tab/name is selected but the body still points to the file tab

Screen Shot 2022-06-14 at 12 54 14 PM
.

app/views/dataset.scala.html
MBcode: tabs v spaces; fixed

MBcode:
Same logic used to make "tab-pane fade in active" along with with the tab having class="active"

lmarini (changes requested):
Please updated changelog. Otherwise good to go. Thank you.

lmarini (approved):

PR #131, opened 2020-11-10
Do not index datasets & collections in trash

max-zilla:
The indexAll functions for collections and datasets were not updated to omit trashed resources, resulting in larger indexing jobs than necessary on instances with large {"trash"=true} documents. This PR omits those trashed documents.


max-zilla:
This still needs more work I think - the File model doesn't have trash in it, so all files in trashed datasets would still be queued here. Marking draft.

lmarini (approved):

PR #137, opened 2020-11-18
Submit marked files in dataset for extraction

max-zilla:
Add a Mark All option for files in a dataset, and Submit option for marked files. Cleared up some naming as well based on feedback. Now the Marked Files dropdown is active even with 0 selections so Mark All is reachable.
Screen Shot 2020-11-18 at 8 19 49 AM

Screen Shot 2020-11-18 at 8 19 42 AM
 
Screen Shot 2020-11-18 at 8 19 57 AM


bodom0015 (commented):
This is a very cool feature! I would recommend hiding the "Submit All" button behind a conditional checking that the RabbitMQ plugin is enabled. With other cases, "Submit to extractor" is hidden if this plugin is not enabled.

Currently, this view still shows when the plugin is disabled. All submissions from the view appear successful and shows a notification, but the messages are never actually sent to RabbitMQ. This caused me a bit of confusion before realizing that it just needed RabbitMQ enabled in `play.plugins`

max-zilla:
Great catch, thanks @bodom0015 .

Updated so now Submit only shows if RMQ is enabled and not if disabled.

lmarini (approved):
Works well! I would like to propose a couple of potential tweaks to the signature of the  endpoint `POST /api/selected/submit/:ds_id/:file_ids ` for a followup PR since we want to get this out. 

I think we should drop `submit`, since it's a verb and our verbs are GET, POST, etc. (We use it in other places in the API, but I am not sure it is a good idea) Also we should try to keep the hierarchical resource uri constraint.

What if we changed this to /api/datasets/:ds_id/files:file_ids/extractions?

This would be inline with POST /api/files/:file_id/extractions and POST /api/datasets/:ds_id/extractions.

Another alternative could be to include a JSON body with information about the dataset and file ids, with the resource being POST /api/extractions. This would avoid having to embed multiple file ids in the URI.

This would also move it outside the /api/selected resource, which we track server side with a different model (versus client side and cookies). This could avoid confusion going forward.

Once this is done we should also add swagger documentation for it. Thanks!

PR #138, opened 2020-11-18
Initial implementation of new event & extractor

max-zilla:
This adds a new clowder.file.submitted and clowder.dataset.submitted event for when those things are manually submitted for extraction. I am still working through required events, so this is Draft for now.

There is also a new extractor (placeholder now) that will capture:
- bytes of files uploaded and deleted by user (tracking storage usage)
- extractions submitted by user with job id, either manually or via upload (tracking extractor usage)


max-zilla:
@lmarini explained that we are moving away from queue bindings, so I'm going to use pattern similar to archive extractor and adding config for an extractor queue you want to send these events to. The MongoDB implementation could eventually just be one flavor.

max-zilla:
https://github.com/clowder-framework/clowder/pull/155/files code moved to this one

PR #140, opened 2020-11-18
Remove swap functionality on reindex

max-zilla:
This will also remove the deletion of the index. This is a hotfix pending a more permanent solution.


lmarini (changes requested):
Can you update the version numbers? I can also do it if you are not familiar with all the files that need updating.

robkooper:
Lazy man: 
```
for x in doc/src/sphinx/conf.py project/Build.scala public/swagger.yml; do sed -i~ 's/1.12.1/1.12.2/g' $x; done
```

robkooper (approved):

lmarini (approved):

PR #144, opened 2020-11-30
editor can now successfully delete tags

tcnichol:
## Description
Delete tags methods in the services for Files and Datasets are modified so that editors in a space can delete tags.

If this is the desired behavior, it's a quick fix.

Otherwise, I can change this and make it so users can clearly only delete tags they added themselves by modifying the Permissions.


app/services/mongodb/MongoDBDatasetService.scala
```diff
@@ -1075,7 +1075,7 @@ class MongoDBDatasetService @Inject() (
   def removeTags(id: UUID, userIdStr: Option[String], eid: Option[String], tags: List[String]) {
```
lmarini, line 1075: Since we are not using `userIdStr: Option[String], eid: Option[String]` should we remove them from the method definition and refactor the code that calls it?

lmarini (changes requested):

max-zilla:
Created a new user without admin permissions, added them as editor to a space and (as admin) added tag to a dataset in it.

New editor user was about to delete tags the admin added. I think this is the intended behavior. With Viewer permissions, new user could not remove.

max-zilla (approved):
need to fix changelog conflict, otherwise good.

app/services/mongodb/MongoDBDatasetService.scala
lmarini: Working on this.

lmarini (approved):

PR #147, opened 2020-12-02
v1.13.0

lmarini:
### Added
- Ability to submit multiple selected files within a dataset to an extractor.
- Support for Amplitude clickstream tracking. See Admin -> Customize to configure Amplitude apikey.
- UpdateUserId.js to scripts/updates. This code adds user_id to each document in extractions collection in mongodb. 
  user_id is taken from author id in uploads.files if exists, else it taken from author id in datasets collection.

### Fixed
- An extractor with file matching set to `*/*` (all file types) would incorrectly send out dataset events.
- Space Editors can now delete tags on files, datasets and sections.
- GeospatialViewer previewer no longer shows if file does not contain geospatial layers.


max-zilla:
Just pushed fix to bug found where files inside folders weren't working properly with Mark All. Changed text to Mark Visible to reflect fixed behavior.

max-zilla (approved):
Tested marking functions, the tag deletion, triggering extractors, but not able to easily test Amplitude tracking or geojson previewer.

PR #152, opened 2020-12-15
Make space ID optional on storage reports

max-zilla:
This makes a few minor changes to make the /api/reports/storage/spaces endpoint not require a space ID but run the report on whole database if none is given.

### DIFF
--- a/public/swagger.yml
+++ b/public/swagger.yml
@@ -5005,17 +5005,17 @@ paths:
         Must be a server admin to access the report. Supports ISO8601 date range filters.
       parameters:
         - name: space
-          in: path
+          in: query
           required: false
           schema:
             type: string
         - name: since
-          in: path
+          in: query
           required: false
           schema:
             type: string
         - name: until
-          in: path
+          in: query
           required: false
           schema:
             type: string
--- a/public/swagger.yml
+++ b/public/swagger.yml
@@ -4995,18 +4995,18 @@ paths:
           description: Not authorized
           content: {}
 
-  /reports/storage/spaces/{id}:
+  /reports/storage/spaces:
     get:
       tags:
         - reporting
       summary: Download storage metrics report for spaces as CSV.
       description: |
-        Download storage metrics report for files in a space as a CSV file.
+        Download storage metrics report for files in a space as a CSV file. Space id is optional parameter.
         Must be a server admin to access the report. Supports ISO8601 date range filters.
       parameters:
-        - name: id
+        - name: space
           in: path
-          required: true
+          required: false
           schema:
             type: string
         - name: since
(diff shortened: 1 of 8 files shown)


bodom0015 (changes requested):
Swagger may be slightly incorrect for passing the space id? Your `swagger.yaml` changes say that `space` is supposed to be provided in the path, but I can't seem to figure out how to pass a space id to this endpoint.

I've tried passing it in the path (`/api/reports/storage/spaces/5f7647058d6a63c38911fc78`), but I think we either need to add `{space}` to the path or change the value of `in` for this parameter. This is the error I get passing the space id via that path:
```json
{"status":"not found","request":"GET /api/reports/storage/spaces/5f7647058d6a63c38911fc78"}
```

I also tried `/api/reports/storage/spaces&space=5f7647058d6a63c38911fc78`, but that yielded the same error:
```json
{"status":"not found","request":"GET /api/reports/storage/spaces&space=5f7647058d6a63c38911fc78"}
```

max-zilla:
> `/api/reports/storage/spaces` seems to work, but Swagger may be slightly incorrect for passing the space id? Your `swagger.yaml` changes say that `space` is supposed to be provided in the path, but I can't seem to figure out how to pass a space id to this endpoint. What is the expected URL/format for passing a space id here?
> 
> I've tried passing it in the path (`/api/reports/storage/spaces/5f7647058d6a63c38911fc78`), but I think we either need to add `{space}` to the path or change the `in: path` for this parameter to something else. This is the error I get passing the space id via that path:
> 
> ```json
> {"status":"not found","request":"GET /api/reports/storage/spaces/5f7647058d6a63c38911fc78"}
> ```
> 
> I also tried `/api/reports/storage/spaces&space=5f7647058d6a63c38911fc78`, but that yielded the same error:
> 
> ```json
> {"status":"not found","request":"GET /api/reports/storage/spaces&space=5f7647058d6a63c38911fc78"}
> ```

The intended syntax was your second one, but I changed & to ?
`/api/reports/storage/spaces?space=5f7647058d6a63c38911fc78`
...let me double-check syntax + docs, thx.

bodom0015:
🤦  you are correct! `/api/reports/storage/spaces?space=5f7647058d6a63c38911fc78` worked like a charm

When using the proper syntax, the endpoint works as expected :+1:

public/swagger.yml
```diff
@@ -4995,18 +4995,18 @@ paths:
...
         - reporting
       summary: Download storage metrics report for spaces as CSV.
       description: |
-        Download storage metrics report for files in a space as a CSV file.
+        Download storage metrics report for files in a space as a CSV file. Space id is optional parameter.
         Must be a server admin to access the report. Supports ISO8601 date range filters.
       parameters:
-        - name: id
+        - name: space
           in: path
```
bodom0015, line 5008: I think this is the expected syntax for denoting query-string parameters:
```suggestion
          in: query
```

lmarini: fixed, thank you @bodom0015

lmarini (approved):

public/swagger.yml
max-zilla: thank you

Issue #12, opened 2020-06-25
Bulk delete files


lmarini:
We currently only have an endpoint to delete one file at a time. Client can potentially make a lot of calls when deleting large collections of files. This would also be helpful in case file deletes fail halfway. Maybe even let the system recover from partial deletes.

The new endpoint would accept list of files IDs in the body of the request.

Note: This exists for Datasets using selections. We could just extend selections to include files.

PR #153, opened 2020-12-20
adding a bulk delete files method

tcnichol:
There was an issue (https://github.com/clowder-framework/clowder/issues/12) about bulk delete files tagged api.

Multiple files can be deleted using the marked feature in datasets, but I noticed that there was still no bulk delete endpoint in the api that fit what was requested in the issue.


lmarini:
@max-zilla can you help me review this one? I believe it is related to some of the batch actions you have implemented on files in a dataset. Thank you.

app/api/Files.scala
```diff
@@ -1623,6 +1616,26 @@ class Files @Inject()(
     }
   }
 
+  def bulkDeleteFiles() = PrivateServerAction (parse.json) {implicit request=>
```
lmarini, line 1619: This action would let anyone logged in call this action and the `removeFile` call doesn't seem to reinforce permissions. We should use something like ```def removeFile(id: UUID) = PermissionAction(Permission.DeleteFile, Some(ResourceRef(ResourceRef.file, id)))``` but I am not sure which permission to use since I don't believe we can pass in a list of resources.

lmarini (changes requested):

app/api/Files.scala
tcnichol: Would another option be to files the FileService removeFile method to take permissions into account? It looks like it's supposed to do that, but just doesn't.

tcnichol: I made a commit that does that.

max-zilla: I implemented the ability to check a list of resources at once a while ago: https://github.com/clowder-framework/clowder/blob/develop/app/api/Permissions.scala#L282 I also added a bulk GET for a list of files: https://github.com/clowder-framework/clowder/blob/develop/app/services/FileService.scala#L104

You can use these in combination like this:
https://github.com/clowder-framework/clowder/blob/develop/app/util/SearchUtils.scala#L302

both of these calls return little objects that tell you what does/doesn't have permission and what was/wasn't found (if you asked for multiple files). Please use this pattern instead. The changes here, you added a new permission check inside the file service but existing uses of that call in api Files have already done the check so now it is checking twice. 
If you remove the permission check in the service, and use the pattern above inside bulkDeleteFiles, it will use existing code and be a bit more efficient.

app/services/mongodb/MongoDBFileService.scala
```diff
@@ -814,67 +816,74 @@ class MongoDBFileService @Inject() (
...
       case Some(file) => {
-        if(!file.isIntermediate){
-          val fileDatasets = datasets.findByFileIdDirectlyContain(file.id)
-          for(fileDataset <- fileDatasets){
-            datasets.removeFile(fileDataset.id, id)
-            if(!file.xmlMetadata.isEmpty){
-              datasets.index(fileDataset.id)
-            }
+        val currentResourceRef = ResourceRef(ResourceRef.file, file.id)
+        val hasPermission = Permission.checkPermission(user.get,Permission.DeleteFile, currentResourceRef)
```
max-zilla, line 820: This call is redundant and has already been done in most cases for this function.

app/api/Files.scala
```diff
@@ -1623,6 +1616,26 @@ class Files @Inject()(
...
+    request.user match {
+      case Some(user) => {
+        val fileIds = request.body.\("fileIds").asOpt[List[String]].getOrElse(List.empty[String])
+        if (fileIds.isEmpty){
+          BadRequest("No file ids supplied")
+        } else {
+          for (fileId <- fileIds){
```
max-zilla, line 1626: you can use a pattern like this to check many file permissions at once https://github.com/clowder-framework/clowder/blob/develop/app/util/SearchUtils.scala#L302

max-zilla (changes requested):

max-zilla:
Also please add entry to swagger.yml for your new endpoint.

app/api/Files.scala
tcnichol: @max-zilla thanks. that will be much better. i'll get this changed and push a new commit later today.

tcnichol: @max-zilla I made the change. Let me know if it looks okay.

conf/routes
```diff
@@ -388,6 +388,7 @@ GET            /api/files/:id/blob
 
 # deprecrated use DELETE
 POST           /api/files/:id/remove                                                    @api.Files.removeFile(id: UUID)
+POST           /api/files/bulkRemove                                                    @api.Files.bulkDeleteFiles()
```
robkooper, line 391: since this is a new API endpoint can you update the swagger documentation as well.

robkooper (changes requested):

conf/routes
tcnichol: I added a new entry in the swagger docs

app/api/Files.scala
lmarini: @tcnichol @max-zilla not sure about the changes to https://github.com/clowder-framework/clowder/blob/develop/app/services/mongodb/MongoDBFileService.scala#L818. I don't see are reference to permissions and the changes in the branch just look like spacing changes? Am I looking in the wrong place? Thank you.

lmarini (changes requested):
Just a question about one of the proposed changes, otherwise the rest looks good. Need to add changelog entry.

app/api/Files.scala
tcnichol: That commit is just spacing changes.

The actual change in checking permissions is in this file:

https://github.com/clowder-framework/clowder/pull/153/files'

line 1627 uses the new checkPermission method that takes in a list of resourceRef. I then removed any permission checks elsewhere.

tcnichol: It looks like I removed the redundant check he mentioned a few commits ago on this one.

lmarini (approved):
It works well. Two followups:

- @tcnichol I have created a followup issue https://github.com/clowder-framework/clowder/issues/209
- @max-zilla can you double check your comment and my comments below regarding "remove the permission check in 818 of MongoDB File Service" and see if we missed something?

PR #154, opened 2021-01-06
Release/1.14

lmarini:
### Added
- Added a new `/api/reports/metrics/extractors` report for summarizing extractor usage by user.

### Changed
- `api/reports/storage/spaces` endpoint now accepts a space parameter for ID rather than requiring a space filter.
- Datasets and collections in the trash are no longer indexed for discovery in search services.


max-zilla (approved):
Ran space report with space filter, extractions report with a time filter, and deleted a file & ran reindex. Everything looked good.

PR #159, opened 2021-01-21
When adding a file to a dataset by URL, prioritize the URL `content-type` header

lmarini:
over the file content type established by looking at the file name extension.

This is related to https://github.com/clowder-framework/clowder/issues/139. We are not sure it fixes that problems all together but it is an improvement over the current behaviour.


max-zilla (approved):
seems to work. can test with a specific file if needed.

PR #163, opened 2021-01-29
Use css `word-wrap: break-word;` where possible to break …

### LINKED ISSUE #160: Text wrapping for long file names
For long file names with no spaces in the name the text doesn't wrap around within the div. 

Screen Shot 2021-01-21 at 12 35 25 PM

Looks like `word-wrap: break-word;` fixes the issue. Will need to apply this in a few places.

Screen Shot 2021-01-21 at 12 35 59 PM

A few places where this needs to be added:
- Listing of files
- File page (main title and title on the right column)

lmarini:
…words that are too long. Fixes #160. 

This issue is obvious with files names that are very long.
Applied to fix to collectios, datasets, spaces titles as well where I saw an issue.
There might be more places where this could be applied.

## Review Time Estimate

- [ ] Immediately
- [x] Within one week
- [ ] When possible

## Types of changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Fixed
 - When adding a file to a dataset by URL, prioritize the URL `content-type` header over the file content type established
   by looking at the file name extension.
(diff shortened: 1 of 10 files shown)


tcnichol:
Looks good. Tested space, collection, dataset and file to make sure long names wrapped correctly.

tcnichol (approved):
Tested. All cases looked fine.

PR #172, opened 2021-02-17
add headers when calling /api/me

robkooper:
Same data is now returned as headers, these can be used when using clowder as an authentication call in nginx/traefik etc.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [x] When possible

## Types of changes

## Checklist:

- [ ] I have signed the CLA


bodom0015 (approved):
LGTM :+1:

We should be able to ingest these directly in the Clowder frontend, or use it in conjunction with something like `oauth2-proxy`.

lmarini (approved):

PR #174, opened 2021-02-19
Allow searching by created date

max-zilla:
Adds a new search tag you can use in API query strings and in the search box.

created:2011-01-01
created<=2012-02-01
created=="2011-01-01T03:22:32:123Z"

should all work. I had to do a little bit of handling to make the lte/gte work as expected, because the created date is a full timestamp in ES and if you only provide YYYY-MM-DD, it defaults to 00:00:00 on the day you provided, which means you still get results from 1am, 2am, etc.... on that date if you said >2011-01-01.


bodom0015 (approved):
LGTM :+1: I had to run a reindex first, then everything worked as expected

Minor side note: I would recommend adding some sort of indicator that a search is in progress. For longer queries, it was difficult to tell if the query was still running or if it had finished with no results.

lmarini (approved):

lmarini:
> LGTM 👍 I had to run a reindex first, then everything worked as expected
> 
> Since other columns/fields use the `==` operator, and this case appears to be the outlier syntax-wise, I agree with using `:` as the example value in the "Search Syntax Help" for `created`.
> 
> Minor side note: I would recommend adding some sort of indicator that a search is in progress. For longer queries, it was difficult to tell if the query was still running or if it had finished with no results.

Created https://github.com/clowder-framework/clowder/issues/183 regarding indicator @bodom0015

Issue #173, opened 2021-02-18
Grid view in dataset, folder and search


hoangnguyen177:
**Is your feature request related to a problem? Please describe.**
Grid view is missing in dataset view, folder view and search. Having grid view allows more images to be shown.  

**Describe the solution you'd like**
Add grid view in space, folder and search. 

**Additional context**
This suggestion has been noted [here](https://opensource.ncsa.illinois.edu/confluence/display/CATS/User+Interface+Improvements). However, I still create this issue for this specific feature request.

PR #176, opened 2021-02-21
Issue 173 grid view

7neyugn7:
## Description
As per [request 173](https://github.com/clowder-framework/clowder/issues/173), grid view is added to dataset, folder and search.

The views are as followed:
gridview_dataset
gridview_folder
gridview_search

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [x] When possible

## Types of changes

## Checklist:

- [x] I have signed the CLA


max-zilla:
Can you take a quick look at the marking files system? During testing, on grid view when I would select a file inside a dataset it would set Marked (2) instead of Marked (1). The list view still seemed to add 1 correctly, but there seems to be some weird behavior there. Can you try to recreate? This is used for bulk file operations.

Screen Shot 2021-03-29 at 2 36 12 PM

A lot of the changes are looking good overall!

lmarini:
@bodom0015 can you help review this? Thanks!

bodom0015 (commented):
Took a look at this PR and the `tile-mode` features seem to work as intended.

@max-zilla previously mentioned a bug with the Marked Files counter that appears to still be present. I can confirm that this behavior does not manifest on `develop` or `master`, and that it now affects both the grid and tile views. I am a bit confused tho, since I can't see where this PR touches that add/remove logic specifically...

And because this PR does not touch that code, I can't seem to make proper suggestions where the fix is needed :/

@lmarini @robkooper should this issue be addressed on this PR before merging? Or could this perhaps be merged and we can follow-up with a fix for this bug?

__________________


Digging into the underlying bug, it looks like the problem comes from the use of string/split to store these IDs as a cookie string. When the cookie is "empty", the value is somehow being stored a pair of double-quotes (`"\"\""`) instead of an empty string `("")`. For example, upon first click the cookie value is set to `"\"\",60c2cbf56aec65b8de9a1775"`, and the leading pair of quotes causes an incorrect length of 2 files.

I suspect that this is because `$.cookie` appears to call `JSON.stringify` in the background on the stored value. Calling `JSON.parse` when reading the cookie back out seems to prevent some of the multi-quote issues, and starts to beg the question of if using a JSON array for the whole list might be preferable to string manipulation?

An implementation like the following would store as a JSON array (make sure to `JSON.parse` in the other 2 spots where the `selected-files-@dataset.id` cookie is read):
```javascript
    function addFileToMarked(file_id) {
        var selected = $.cookie('selected-files-@dataset.id');
        if (selected) {
            // Parse JSON array and add new id
            var fileUUIDs = JSON.parse(selected);
            if (fileUUIDs.indexOf(file_id) === -1) {
                fileUUIDs.push(file_id);
            }
            // Set modified array as cookie
            $.cookie('selected-files-@dataset.id', fileUUIDs);
        } else {
            // Set array containing single id as cookie
            $.cookie('selected-files-@dataset.id', [file_id]);
        }

        $("a[data-id='"+file_id+"'] span.glyphicon").removeClass('glyphicon-plus');
        $("a[data-id='"+file_id+"'] span.glyphicon").addClass('glyphicon-ok');
        updateSelectedFileCount();
    }

    function removeFileFromMarked(file_id) {
        var selected = $.cookie('selected-files-@dataset.id');
        if (selected) {
            // Parse JSON array and attempt to remove id
            var fileUUIDs = JSON.parse(selected);
            var index = fileUUIDs.indexOf(file_id);
            if (index !== -1) {
                // if possible, remove the id (only set the cookie if it was modified)
                fileUUIDs.splice(index, 1);
                $.cookie('selected-files-@dataset.id', fileUUIDs, {expires: 14});
            }
        }

        $("a[data-id='"+file_id+"'] span.glyphicon").removeClass('glyphicon-ok');
        $("a[data-id='"+file_id+"'] span.glyphicon").addClass('glyphicon-plus');
        updateSelectedFileCount();
    }

```

PR #191, opened 2021-03-11
Release/1.15.1

lmarini:
### Fixed
- Several views were throwing errors trying to access a None value in `EventSinkService` when a user was not logged in. 
  Replaced `get()` with `getOrElse()`.
- Consolidated field names sent by the EventSinkService to maximize reuse.
- Changed `EventSinkService` logging to debug to minimize chatter.
- Don't automatically create eventsink queue and bind it to eventsink exchange. Let clients do that so that we don't 
  have a queue for the eventsink filling up if there are no consumers.

### DIFF
--- a/app/services/EventSinkService.scala
+++ b/app/services/EventSinkService.scala
@@ -80,8 +80,8 @@ class EventSinkService {
       "resource_name" -> dataset.name,
       "author_id" -> dataset.author.id,
       "author_name" -> dataset.author.fullName,
-      "viewer_id" -> viewer.get.id,
-      "viewer_name" -> viewer.get.getMiniUser.fullName
+      "viewer_id" -> viewer.getOrElse(User.anonymous).id,
+      "viewer_name" -> viewer.getOrElse(User.anonymous).getMiniUser.fullName
     ))
   }
 
@@ -94,8 +94,8 @@ class EventSinkService {
       "resource_name" -> file.filename,
       "author_id" -> file.author.id,
       "author_name" -> file.author.fullName,
-      "viewer_id" -> viewer.get.id,
-      "viewer_name" -> viewer.get.getMiniUser.fullName
+      "viewer_id" -> viewer.getOrElse(User.anonymous).id,
+      "viewer_name" -> viewer.getOrElse(User.anonymous).getMiniUser.fullName
     ))
   }
 
@@ -108,8 +108,8 @@ class EventSinkService {
       "resource_name" -> collection.name,
       "author_id" -> collection.author.id,
       "author_name" -> collection.author.fullName,
-      "viewer_id" -> viewer.get.id,
-      "viewer_name" -> viewer.get.getMiniUser.fullName
+      "viewer_id" -> viewer.getOrElse(User.anonymous).id,
+      "viewer_name" -> viewer.getOrElse(User.anonymous).getMiniUser.fullName
     ))
   }
 
@@ -135,8 +135,8 @@ class EventSinkService {
           "resource_name" -> space.name,
           "author_id" -> author.id,
           "author_name" -> author.fullName,
-          "viewer_id" -> "",
-          "viewer_name" -> "Anonymous"
+          "viewer_id" -> User.anonymous.id,
+          "viewer_name" -> User.anonymous.fullName
         ))
       }
       case (Some(v), None) => {
@@ -159,8 +159,8 @@ class EventSinkService {
           "resource_name" -> space.name,
           "author_id" -> space.creator.stringify,
           "author_name" -> "",
-          "viewer_id" -> "",
-          "viewer_name" -> "Anonymous"
+          "viewer_id" -> User.anonymous.id,
+          "viewer_name" -> User.anonymous.fullName
         ))
       }
     }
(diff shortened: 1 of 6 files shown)


app/services/EventSinkService.scala
```diff
@@ -6,33 +6,21 @@ import java.net.URI
 import java.time.Instant
 import play.api.{Logger, Play}
 import play.api.Play.current
-import play.api.libs.json.{JsValue, Json}
+import play.api.libs.json.{JsObject, JsValue, Json}
 
 object EventSinkService {
   val EXCHANGE_NAME_CONFIG_KEY = "eventsink.exchangename"
   val QUEUE_NAME_CONFIG_KEY = "eventsink.queuename"
```
robkooper, line 13: I believe we can get rid of the queue all together here, and just send to the exchange

app/services/EventSinkService.scala
```diff
@@ -41,20 +29,18 @@ class EventSinkService {
...
-
-    Logger.info("Submitting message to event sink exchange: " + Json.stringify(metadata))
-
-    //val message = EventSinkMessage(Instant.now().getEpochSecond, category, metadata)
-    messageService.submit(exchangeName, queueName, metadata, "fanout")
+  def logEvent(message: JsValue) = {
+    // Inject timestamp before logging the event
+    val event = message.as[JsObject] + ("created" -> Json.toJson(java.util.Date.from(Instant.now())))
+    Logger.info("Submitting message to event sink exchange: " + Json.stringify(event))
+    messageService.submit(exchangeName, queueName, event, "fanout")
```
robkooper, line 36: I think if we pass `""` or None here (not sure what the default is) it will not send it to a queue, but just to the exchange.

lmarini: The queueName is used as routing key further below 

```tempChannel.basicPublish(exchange, routing_key, null, message.toString.getBytes)```

It seemed fine when I ran the code with and without the queue.

**Review by max-zilla — APPROVED** [review]
Ran and verified new category field for events, didn't encounter issues. Didn't catch anything in code.

File: app/services/rabbitmq/RabbitMQMessageService.scala [hunk 3]
```diff
@@ -233,8 +233,6 @@ class RabbitMQMessageService extends MessageService {
     // This probably isn't going to extract queue (use other submit() for that) so make a new broker
     val tempChannel = connection.get.createChannel()
     tempChannel.exchangeDeclare(exchange, exchange_type, true)
```
bodom0015, line 235:
```suggestion
    tempChannel.exchangeDeclare(exchange, exchange_type, true)
    
    // If a routing_key (queue name) was provided, ensure that the queue exists
    if (routing_key != "") {
      tempChannel.queueDeclare(routing_key, true, false, false, null)
      tempChannel.queueBind(routing_key, exchange, routing_key)
    }
```

I think we should either remove the `routing_key` parameter for this method, or properly handle both cases here. If a user passes an empty routing_key, we send to the exchange. Otherwise, we make sure the target queue exists before publishing. This way, we can change the default to "" in `EventSinkService` if we prefer the exchange-based approach, while still preserving functionality for a non-exchange based approach to sending RMQ messages.

bodom0015 (changes requested):
RabbitMQMessageService.submit(exchange: String, routing_key: String, message: JsValue, exchange_type: String = "topic") now accepts a `routing_key` whose value may not be valid.

app/services/EventSinkService.scala
```diff
@@ -6,33 +6,21 @@ import java.net.URI
...
 import play.api.Play.current
-import play.api.libs.json.{JsValue, Json}
+import play.api.libs.json.{JsObject, JsValue, Json}
 
 object EventSinkService {
   val EXCHANGE_NAME_CONFIG_KEY = "eventsink.exchangename"
   val QUEUE_NAME_CONFIG_KEY = "eventsink.queuename"
 
   val EXCHANGE_NAME_DEFAULT_VALUE = "clowder.metrics"
   val QUEUE_NAME_DEFAULT_VALUE = "event.sink"
```
bodom0015, line 16:
```suggestion
  val QUEUE_NAME_DEFAULT_VALUE = ""
```

Set default queue name to "" to send to entire exchange by default

app/services/EventSinkService.scala
lmarini: We will need to revisit this call. We are currently using the name of the queue as both the queue name and routing key. The meaning of the routing key for exchange is deepened on the the type of exchange. For example, in the current use of this function we are using fanout and that one doesn't use routing keys.

bodom0015, line 36:
```suggestion
    try {
      messageService.submit(exchangeName, queueName, event, "fanout")
    } catch {
      case e: Throwable => { Logger.error("Failed to submit event sink message", e) }
    }
```

Giving this one final look, I'm noticing that the user can't log in if RabbitMQ is not running. The exception doesn't cause Clowder to crash, but prevents a log in simply because the event cannot be logged.

If RabbitMQ is still an optional dependency, we might consider catching these exceptions with something like the above.

bodom0015 (approved):
LGTM, aside from the issue noted below. If we're moving RabbitMQ to a hard requirement then please disregard, but I don't recall that being part of the discussion of these feature.

app/services/EventSinkService.scala
lmarini: if MessageService is not wired to Rabbitmq it will call the `noop` function which will log something. I think the pattern is to add a call to `connect()` in every Rabbitmq call that makes sure it is connected to the broker (see other function bodies in the Rabbitmq implementation). Can we add that to this `submit` implementation?

No matter what, I am not sure what will happen if `connect()` fails. It might still fail and this catch would still be needed.

PR #195, opened 2021-03-17
Add file status to file report

max-zilla:
This adds a status column to the files admin report to indicate if a file is ARCHIVED, etc.


lmarini (changes requested):
It's putting the status in the `parent_spaces` column for me. Quickly looking at the code it looks like it might not throw an exception and so it skips a column?

Also it show `"PROCESSING` maybe one quote is printed out the the other is not.


Screen Shot 2021-03-29 at 8 38 02 AM

lmarini (approved):

PR #204, opened 2021-04-02
Add geoserver yml

tcnichol:
With this pull request, I have added a .yml file for the geoserver
next,

docker-compose -f docker-compose.yml -f docker-compose.geoserver.yml  up -d

This will run clowder with geoserver and the extractors.

I am making this a work in progress and asking for feedback for the following reasons.

1. I have not yet figured out how to use the proxy server.

Feel free to add any feedback.


docker-compose.geoserver.yml
```diff
@@ -0,0 +1,37 @@
...
+      - clowder
+#    user: "55242:17027"
+    environment:
+      - RABBITMQ_URI=${RABBITMQ_URI:-amqp://guest:guest@rabbitmq/%2F}
+      - RABBITMQ_EXCHANGE=${RABBITMQ_EXCHANGE:-clowder}
+      - REGISTRATION_ENDPOINTS=${REGISTRATION_ENDPOINTS}
+      - PROXY_ON=${PROXY_ON}
+      - PROXY_HOST=${PROXY_HOST}
+      - PROXY_URL=${PROXY_URL}
+      - GEOSERVER_USERNAME=${GEOSERVER_USERNAME:-admin}
```
robkooper, line 17: can you add these to the env.example and document them

tcnichol: Yes, can do.

tcnichol:
NOTE 

This works with docker-compose, but I have not been able to figure out how to get the proxy to work. 

My plan is to have this pull request, and then create a new issue for the proxy, and do that on a later pull request. 

As things are now, they at least can be used with docker-compose.

max-zilla:
@tcnichol looks like env.example needs to be updated still, can you add those and then mark this not as draft? otherwise looks good.

max-zilla (changes requested):
update env.example

tcnichol:
I updated the env.example

Let me know if there are better values to put in than what I have there.

max-zilla (approved):

PR #216, opened 2021-05-05
Persistent status queue

robkooper:
## Description

When (re)starting clowder it will create an anonymous queue for responses. This will make sure clowder has a persistent ID that is used for the response queue for extractors. This will make it such that if clowder restarts it will get all the status messages from extractors.

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [x] When possible

## Types of changes

## Checklist:

- [ ] I have signed the CLA


lmarini (approved):
- should we include `instance` in `application.config`? Otherwise every times it restart it will be a different number?
- should we rename instance to instanceId or something similar?

PR #226, opened 2021-05-11
use clowder.url for absolute url (fixes #224)

### LINKED ISSUE #224: Ability to set clowder url
When enable the auto-archive we need to be able to set the clowder.url. One option is to set clowder.rabbitmq.clowderurl, but other places we might want to be able to set the absolute clowder url.

robkooper:
## Description

This will allow you to set the URL for clowder. Anything that uses a absolute url will use this configuration option if available.

To test set in your custom.conf:
```
clowder.url=http://clowder.example.com:9000
```

And add to /etc/hosts:
```
127.0.0.1 clowder.example.com
```

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [X] When possible

## Types of changes

- [X] New feature (non-breaking change which adds functionality)

## Checklist:

- [ ] I have signed the CLA

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## Unreleased
+
+### Added
+- Can set the clowder url in configuration file [#224](https://github.com/clowder-framework/clowder/issues/224)
+
 ## 1.17.0 - 2021-04-29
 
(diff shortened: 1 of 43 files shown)


robkooper:
Thinking about this, not sure I like the result of it.

If you access it with one url, it will always try to redirect you to an absolute url in certain cases. I think it would be better to remove all absolute URL's where we can.

robkooper:
going to close this PR, this does not do what I really wanted to do. What I tried to solve is the ability to specify the URL that is the prefix, but if you have multiple servers hostnames pointing to the same clowder instance it becomes a mess.

Issue #34, opened 2020-07-17
Include folder name and ID in dataset file lists


max-zilla:
/api/datasets/:id/files will list all files in the dataset as JSON objects, but files within folders have no indication they are not at the root level, making certain operations difficult.

for each file:
```
{
size: "657",
date-created: "Wed Jul 08 10:25:09 CDT 2020",
id: "5f05e555066c371087387c78",
contentType: "text/plain",
filename: "tempnotes.txt"
}
```

I would propose:
```
{
size: "657",
date-created: "Wed Jul 08 10:25:09 CDT 2020",
id: "5f05e555066c371087387c78",
contentType: "text/plain",
filename: "tempnotes.txt",
folder: {
  name: "MyFolder",
  id: "438904901412412412"
}
}
```

lbcsept:
While waiting for this issue to be solved, is there a way to get the folder that contains a specific file, or the reverse : is that possible to get the list of files within a specific folder_id ?

tcnichol:
I created a pull request that I believe fixes this issue.

lmarini:
Thanks @tcnichol !

PR #229, opened 2021-05-15
new method adds folder id and name to file in datasets files endpoint

tcnichol:
## Description
This fixes issue https://github.com/clowder-framework/clowder/issues/34

The folder name and id are now included in the json for api/dataset/:id/files

New method added for getting jsonFilesWithFolder in api.Datasets


app/api/Datasets.scala
```diff
@@ -1134,22 +1128,56 @@ class  Datasets @Inject()(
...
           })
+          print("done with folder")
         }
       }
       case None => Logger.error(s"Error getting dataset $id")
     }
     output.toList
   }
 
+  def jsonFileWithFolder(file: models.File, folder: models.Folder, serverAdmin: Boolean = false) : JsValue = {
```
robkooper, line 1150: since this is pretty much a copy/paste from below, why not add tit to below as option with None as default?

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
+- Added folder and folder id to api datasets files list [#34](https://github.com/clowder-framework/clowder/issues/34)
```
robkooper, line 8: can you update swagger and add the returned json document to the correct function.

robkooper (changes requested):

app/api/Datasets.scala
tcnichol: done

CHANGELOG.md
tcnichol: i'm seeing related methods starting on line

2643 
and
2530.

The response just says '200' and 'ok.' This is in the swagger.yml. Am I looking in the wrong place to change this?

lmarini: @tcnichol Rob is talking about the return type. We don't do this for most of the endpoints. I made a new task for doing that https://github.com/clowder-framework/clowder/issues/247. Since we want to merge this today I think we are good without it (I tried to do it but ran out of time and didn't want to accidentally break the docs in a hurry).

lmarini (approved):

PR #230, opened 2021-05-17
209 bulk delete permissions

tcnichol:
The bulk delete files endpoint is now changed. Instead of returning just 'ok' it now returns a json containig

list of ids deleted
list of ids not found
list of ids no permission to delete
list of ids error deleting.


tcnichol:
I'm sending a json now that will list file ids that were deleted, not found, not deleted due to permission or error deleting.

lmarini (approved):

PR #233, opened 2021-05-21
232 upload file to folder

tcnichol:
This pull request enabled users to upload files to a folder in a dataset through the api. Previously this option was not available.

The route api.Files.uploadToDataset now has a parameter folder_id: Option[String]. 

Swagger documentation for that route has also been updated.


app/api/Files.scala
```diff
@@ -523,14 +523,47 @@ class Files @Inject()(
   /**
     * Upload a file to a specific dataset
     */
-  def uploadToDataset(dataset_id: UUID, showPreviews: String = "DatasetLevel", originalZipFile: String = "", flagsFromPrevious: String = "", extract: Boolean = true) = PermissionAction(Permission.AddResourceToDataset, Some(ResourceRef(ResourceRef.dataset, dataset_id)))(parse.multipartFormData) { implicit request =>
+  def uploadToDataset(dataset_id: UUID, showPreviews: String = "DatasetLevel", originalZipFile: String = "", flagsFromPrevious: String = "", extract: Boolean = true, folder_id: Option[String] = None) = PermissionAction(Permission.AddResourceToDataset, Some(ResourceRef(ResourceRef.dataset, dataset_id)))(parse.multipartFormData) { implicit request =>
```
robkooper, line 526: would it be better to use Option[UUID] instead for the folder, that way we know it is a UUD.

app/api/Files.scala
```diff
@@ -523,14 +523,47 @@ class Files @Inject()(
...
     datasets.get(dataset_id) match {
       case Some(dataset) => {
-        val uploadedFiles = FileUtils.uploadFilesMultipart(request, Some(dataset), showPreviews = showPreviews, originalZipFile = originalZipFile, flagsFromPrevious = flagsFromPrevious, runExtractors = extract, apiKey = request.apiKey)
-        uploadedFiles.length match {
-          case 0 => BadRequest("No files uploaded")
-          case 1 => Ok(Json.obj("id" -> uploadedFiles.head.id))
-          case _ => Ok(Json.obj("ids" -> uploadedFiles.toList))
+        folder_id match {
```
robkooper, line 529: I think we can simplify (not tested)
```
val folder = folder_id.flatmap(x => folders.get(x))
val uploadedFiles = FileUtils.uploadFilesMultipart(request, Some(dataset), folder, showPreviews = showPreviews, originalZipFile = originalZipFile, flagsFromPrevious = flagsFromPrevious, runExtractors = extract, apiKey = request.apiKey) 
uploadedFiles.length match {
    case 0 => BadRequest("No files uploaded")
    case 1 => Ok(Json.obj("id" -> uploadedFiles.head.id))
    case _ => Ok(Json.obj("ids" -> uploadedFiles.toList))
}
```
This will use flatmap (if there is a value, return an option) to get the folder and then just call the next function with ether None, or Some(folder).

robkooper (changes requested):

app/api/Files.scala
tcnichol: Fixed.

app/api/Files.scala
tcnichol: This works.

However, I have another problem.

I have been testing this with the following python script 

https://github.com/tcnichol/scripts/blob/master/upload_with_folder.py

The method 'upload_a_file_to_dataset_with_folder' works. 

Before I added the folder_id parameter, the method 'upload_a_file_to_dataset' worked. Now it no longer does; I get a 500 error unless I supply a value for folder_id.

But since folder_id is not a required parameter and has a default value, I am not sure why this is happening. Obviously before the same type of call worked but without any optional parameters being supplied. 

I am going to try to add the folder_id to the multipart form data.

tcnichol: I am now using a conditional that seems to handle all cases without breaking. 

Regrettably it kept giving me an error if I tried to put an Option[UUID] in the api route and didn't send it, but this seems to match what is done in a few other cases with optional uuid parameters.

app/api/Files.scala
```diff
@@ -523,10 +523,18 @@ class Files @Inject()(
   /**
     * Upload a file to a specific dataset
     */
-  def uploadToDataset(dataset_id: UUID, showPreviews: String = "DatasetLevel", originalZipFile: String = "", flagsFromPrevious: String = "", extract: Boolean = true) = PermissionAction(Permission.AddResourceToDataset, Some(ResourceRef(ResourceRef.dataset, dataset_id)))(parse.multipartFormData) { implicit request =>
+  def uploadToDataset(dataset_id: UUID, showPreviews: String = "DatasetLevel", originalZipFile: String = "", flagsFromPrevious: String = "", extract: Boolean = true, folder_id: Option[String]) = PermissionAction(Permission.AddResourceToDataset, Some(ResourceRef(ResourceRef.dataset, dataset_id)))(parse.multipartFormData) { implicit request =>
     datasets.get(dataset_id) match {
       case Some(dataset) => {
-        val uploadedFiles = FileUtils.uploadFilesMultipart(request, Some(dataset), showPreviews = showPreviews, originalZipFile = originalZipFile, flagsFromPrevious = flagsFromPrevious, runExtractors = extract, apiKey = request.apiKey)
+
```
robkooper, line 529: no need to change code, stylistic only 1 new line is needed here and maybe add a space between ) and { below, but that is nitpicking code style.

robkooper (approved):

app/api/Files.scala
tcnichol: Fixed.

app/api/Files.scala
```diff
@@ -523,10 +523,15 @@ class Files @Inject()(
   /**
     * Upload a file to a specific dataset
     */
-  def uploadToDataset(dataset_id: UUID, showPreviews: String = "DatasetLevel", originalZipFile: String = "", flagsFromPrevious: String = "", extract: Boolean = true) = PermissionAction(Permission.AddResourceToDataset, Some(ResourceRef(ResourceRef.dataset, dataset_id)))(parse.multipartFormData) { implicit request =>
+  def uploadToDataset(dataset_id: UUID, showPreviews: String = "DatasetLevel", originalZipFile: String = "", flagsFromPrevious: String = "", extract: Boolean = true, folder_id: Option[String]) = PermissionAction(Permission.AddResourceToDataset, Some(ResourceRef(ResourceRef.dataset, dataset_id)))(parse.multipartFormData) { implicit request =>
     datasets.get(dataset_id) match {
       case Some(dataset) => {
-        val uploadedFiles = FileUtils.uploadFilesMultipart(request, Some(dataset), showPreviews = showPreviews, originalZipFile = originalZipFile, flagsFromPrevious = flagsFromPrevious, runExtractors = extract, apiKey = request.apiKey)
+        val current_folder = if (UUID.isValid(folder_id.get)){
```
lmarini, line 529: This throws an error if `folder == None`

lmarini (changes requested):
(This PR is not urgent, it can go in the next release) There is one issue mentioned below. Can you also please updated the changelog? Thank you!

app/api/Files.scala
tcnichol: I believe I have fixed this. Also updated changelog

max-zilla (approved):

PR #234, opened 2021-05-24
Ability to queue archive / unarchive for full datasets

bodom0015:
### Problem
The new auto-archiving feature is very nice, but adds the possibility that many files could be accidentally archived en masse. Unarchiving, in this case, would be very tedious as each file needs to be unarchived individually, which requires a lot of UI navigation, clicking, and waiting.

### Approach
Save the user in this case by offering an "Archive All" and "Unarchive All" button at the dataset-level

These buttons will quickly recurse over the contents of the dataset and queue each file inside for archival / unarchival respectively.

I've also added a new Permission for `ArchiveDataset` (since we previously added `ArchiveFile`, we should match that pattern)

### How to Test
Prerequisite: Minio running, Clowder configured to use `S3ByteStorageDriver`, S3 archival extractor configured and running

1. Checkout and run this branch locally
2. Login to Clowder
3. Create a dataset, if you haven't already
4. Within the dataset, upload a file at the root level and create a folder
5. Navigate into the folder you just created, upload a file, and create another folder
6. Navigate once again into this folder and upload one last file
    * You should now have a dataset that looks roughly as follows:
        * Dataset
            * Folder A
                * Folder B
                   * File 3
                * File 2
            * File 1

## Review Time Estimate

- [ ] Immediately
- [x] Within one week
- [ ] When possible

## Types of changes

## Checklist:

- [ ] I have signed the CLA


app/api/Files.scala
```diff
@@ -1932,7 +1932,7 @@ class Files @Inject()(
       }
       case None => {
         Logger.error("Error getting file " + id)
-        InternalServerError
+        NotFound
```
robkooper, line 1935: this is a good change, thanks for fixing this.

robkooper:
To fix https://github.com/clowder-framework/extractors-archival/issues/5 can you change the `last_donwloaded` to be now()

bodom0015:
@robkooper this was tricky - since we don't expose a method to directly update the `last_download` date, I've added an optional path to `incrementDownloads`. By default this method will increment both date and count (preserving the current behavior that we see today), but I've added an optional `dateOnly` parameter (false by default) here to allow us to update only the Date of the last download.

This seemed preferable to creating an entire new method that would duplicate half of the logic already present here.

lmarini (approved):

PR #241, opened 2021-06-11
Adds MRI previewer for NIFTY (.nii) file format.

pgroves:
## Description

Adds an MRI previewer that is being used on the PigBrainMRI project with Dilger's lab. See the README.md in the previewer directory for details.

## Review Time Estimate

- [ ] Immediately
- [x ] Within one week
- [ ] When possible

## Types of changes

- [ x] New feature (non-breaking change which adds functionality)

## Checklist:

- [ ] I have signed the CLA


pgroves:
There is a problem ... since I moved the previewer from custom/public/ into public/ the gzipped versions of the images don't work (the previewer doesn't load at all). On my deployment where the previewer is in custom/, a file with extension .nii.gz loads the previewer as expected, but nothing happens on the deployment I'm testing this PR with.  I don't know how to debug such a thing... the javascript console doesn't show any errors.

lmarini:
> There is a problem ... since I moved the previewer from custom/public/ into public/ the gzipped versions of the images don't work (the previewer doesn't load at all). On my deployment where the previewer is in custom/, a file with extension .nii.gz loads the previewer as expected, but nothing happens on the deployment I'm testing this PR with. I don't know how to debug such a thing... the javascript console doesn't show any errors.

I have the same problem. I manually unzipped them and it worked fine. What do you have in your custom mimetypes.conf on your deployment? In this pr you have only included `mimetype.nii=image/nii`. I tried adding `mimetype.nii.gz=image/nii` but it doesn't like it because of the extra fragments.

pgroves:
> > There is a problem ... since I moved the previewer from custom/public/ into public/ the gzipped versions of the images don't work (the previewer doesn't load at all). On my deployment where the previewer is in custom/, a file with extension .nii.gz loads the previewer as expected, but nothing happens on the deployment I'm testing this PR with. I don't know how to debug such a thing... the javascript console doesn't show any errors.
> 
> I have the same problem. I manually unzipped them and it worked fine. What do you have in your custom mimetypes.conf on your deployment? In this pr you have only included `mimetype.nii=image/nii`. I tried adding `mimetype.nii.gz=image/nii` but it doesn't like it because of the extra fragments.

I don't see anything else with mimetypes in the working version in custom. You can see it all here:
[https://github.com/ncsa/clowder-dilger-ops/tree/main/custom](https://github.com/ncsa/clowder-dilger-ops/tree/main/custom)

The only other thing I can think of is that the working version is using the official docker images but I rebuilt the clowder image to test this  (from whatever was on master yesterday).

FYI here is what it looks like in the working version, the mimetype is just `application/gzip`. Reading a little about mimetypes this seems ok b/c gzip is the encoding, not really the mimetype, and the mimetype ecosystem seems to treat zips as a special case. But obviously it all depends on the library you're using. 
gzipped_nii_working

lmarini (approved):
I added application/gzip to the file types it triggers on and everything worked well with my test example. Also added changelog and your name to contributing. Thank you!

PR #243, opened 2021-06-14
storage fixes

robkooper:
If storage was enabled in .env it was not reflected correctly in the docker-compose file

add storage information to api/status

image

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [X] When possible

## Types of changes

## Checklist:

- [ ] I have signed the CLA


lmarini (changes requested):
This will show some internal info, like the path on disk to anyone not logged in. Do we want to do that? I was thinking we might go beyond even showing it to users logged in and only show it to users who are in superadmin mode. What do you think?

lmarini (approved):

PR #267, opened 2021-08-18
266 add when parameter in api datasets, collections and spaces

longshuicy:
## Description
add "when" parameter in below endpoints:
```
GET /api/datasets
GET /api/datasets/canEdit

GET /api/collections
GET /api/collections/canEdit  
GET /api/collections/possibleParents
GET /api/collections/{coll_id}/datasetPossibleParents/{ds_id}
GET /api/collections/list 

GET /api/spaces
GET /api/spaces/canEdit
```
## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [x] When possible

## Types of changes

## Checklist:

- [ ] I have signed the CLA

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## Unreleased
+### Added
+- Add "when" parameter in a few GET API endpoints to enable pagination [#266](https://github.com/clowder-framework/clowder/issues/266)
+
+
 ## 1.18.1 - 2021-08-16
 
 This release fixes a critical issue where invalid zip files could result in the files not being uploaded correctly. To check to see if you are affected, please use the following query:
--- a/app/api/Collections.scala
+++ b/app/api/Collections.scala
@@ -237,15 +237,15 @@ class Collections @Inject() (datasets: DatasetService,
     }
   }
 
-  def list(title: Option[String], date: Option[String], limit: Int, exact: Boolean) = PrivateServerAction { implicit request =>
-    Ok(toJson(listCollections(title, date, limit, Set[Permission](Permission.ViewCollection), false, request.user, request.user.fold(false)(_.superAdminMode), exact)))
+  def list(when: Option[String], title: Option[String], date: Option[String], limit: Int, exact: Boolean) = PrivateServerAction { implicit request =>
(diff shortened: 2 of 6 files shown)


bodom0015:
@longshuicy so it looks like the `when` parameter can take either "a" (after) or "b" (before) as a value, is that right? Do you happen to have any examples of a few queries that are expected to succeed?

@robkooper @lmarini I'm also having trouble figuring out the correct format for the `date` parameter. Looking through the code, shouldn't this be an ISO 8601-like timestamp?

 Would recommend adding the related parameters to `public/swagger.yaml` to simplify testing the new API endpoints:
```yaml
      parameters:
        - name: when
          in: query
          required: false
          schema:
            type: string
        - name: date
          in: query
          required: false
          schema:
            type: string
        - name: title
          in: query
          required: false
          schema:
            type: string
```

robkooper:
@bodom0015 thanks for catching the swagger docs update.

longshuicy:
@bodom0015 
yes, it takes either a (after) or b (before). but only "a" is enforced, other characters even though it's not "b" is counted as nextPage = false.

I copied that logic from the controller methods. e.g. GET /datasets --> 
```
def list(when: String, date: String, limit: Int, space: Option[String], status: Option[String], mode: String, owner: Option[String], showPublic: Boolean, showOnlyShared : Boolean, showTrash : Boolean) = UserAction(needActive=false) { implicit request =>
    implicit val user = request.user

    val nextPage = (when == "a")
    val person = owner.flatMap(o => users.get(UUID(o)))
    val ownerName = person match {
      case Some(p) => Some(p.fullName)
      case None => None
    }
...
...
```

longshuicy:
@bodom0015 here is what i use to test
http://localhost:9000/api/datasets?superAdmin=true&limit=3&when=b&date=2021-07-28T16:24:49.000Z
hmm you might be missing a Z in the end

longshuicy:
updated the swagger. Thanks! @bodom0015

lmarini:
please update changelog

bodom0015 (approved):
@longshuicy thank you, the missing Z was indeed the problem there!

I think my confusion here was that the milliseconds **and** timezone are both required.

For example, despite them all describing the same epoch moment:
* Invalid: `2021-05-31T16:33:54+00:00` and `2021-05-31T16:33:54Z` 
* Valid: `2021-05-31T16:33:54.000+00:00` and `2021-05-31T16:33:54.000Z`

Either way, the format stuff is outside of the scope of this PR! :+1: LGTM

max-zilla (approved):

PR #273, opened 2021-09-15
adding geojson mime type

tcnichol:
Very simple pull request.

Just adds geojson as a mime type.

There should be no error on geojson files now, as they won't be used in the previewer.


lmarini:
Please add changelog. Thanks!

longshuicy (approved):

PR #321, opened 2022-02-01
don't show event stream even when logged in (fixes #280)

### LINKED ISSUE #280: disable events does not disable events when user logged in
The disable events flag does not disable events when the user is logged in. That value is only checked when there is no logged in user.

If you set the value of  clowder.disable.events=true you will still see events on the page.

robkooper:
when the user is logged in, it still showed the event stream. This now disables event stream

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [X] When possible

## Types of changes

## Checklist:

- [ ] I have signed the CLA


lmarini (approved):

tcnichol (approved):
I will mark this approved. Works, all changes seem right.

PR #332, opened 2022-03-02
file/dataset add schema.org

MBcode:
## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [x] When possible

## Types of changes

## Checklist:

- [x] I have signed the CLA


MBcode:
It still needs a little formatting, eg. for authour..
and
I would like to add @id's for every page's json-ld
 then in the 'dataset', I can just ref the @id of the 'file's jsonLD
 even though the sitemap(coming soon), usually only go to those 1st dataset pages
So I might bring it up into the dataset page

I had this under a branch called schema.org, but this 1st cut only has metadata from those 2 classes
So am calling it: class2SOmetadata, later we can consider mapping elts from the metadata editor
and just using any attached uploaded jsonld

MBcode:
No sitemap generation in this go, but do have File and Dataset attributes mapped to schema.org predicates and embedded in each of those pages as ld+json that can be parsed by https://validator.schema.org

lmarini (changes requested):

app/views/dataset.scala.html
```diff
@@ -382,6 +382,24 @@ <h4>Statistics</h4>
...
     <script src="@routes.Assets.at("javascripts/people.js")" type="text/javascript"></script>
     <script src="@routes.Assets.at("javascripts/creators.js")" type="text/javascript"></script>
+
+    <script type="application/ld+json">{"context": { "@@vocab": "https://schema.org/"},
+		 "identifier": "@dataset.id",
+		 "name": "@dataset.name",
+		 "author": @Html(dataset.author.to_jsonld()),
+		 "description": "@dataset.description",
+		 "dateCreated": "@dataset.created.format("MMM dd, yyyy")",
+		 "DigitalDocument": "@dataset.files",
```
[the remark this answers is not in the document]
MBcode: LicenseData.scala's urlViaAttributes is trying to get a url for that attribute, by guessing from attributes, but if it doesn't get anything I can have it default to the associated URL.  
I'm also going to join the list of files with commas between them. Thanks! for catching these

app/models/User.scala
```diff
@@ -108,7 +110,28 @@ case class MiniUser(
...
+      /**
+       * return MiniUser as string in jsonld format, w/fullName split into first and last
+       */
+       def to_jsonld () : String = {
+          var firstName= "";
+          var lastName= "";
+          if(fullName.split("\\w+").length>1){
+             lastName = fullName.substring(fullName.lastIndexOf(" ")+1);
+             firstName = fullName.substring(0, fullName.lastIndexOf(' '));
+           } else{ firstName = fullName; }
```
robkooper, line 123:
```suggestion
           } else {
              firstName = fullName;
           }
```

app/models/User.scala
```diff
@@ -108,7 +110,28 @@ case class MiniUser(
...
+	   val authorLD = JsObject(Seq(
+                   "@type" -> JsString("Person"),
+                   "givenName" -> JsString(firstName.toString()),
```
robkooper, line 126: can we use `name` and their fullName

app/views/dataset.scala.html
```diff
@@ -382,6 +382,24 @@ <h4>Statistics</h4>
     <script src="@routes.Assets.at("javascripts/deleteUtils.js")" type="text/javascript"></script>
     <script src="@routes.Assets.at("javascripts/people.js")" type="text/javascript"></script>
     <script src="@routes.Assets.at("javascripts/creators.js")" type="text/javascript"></script>
+
+    <script type="application/ld+json">{"context": { "@@vocab": "https://schema.org/"},
+		 "identifier": "@dataset.id",
+		 "name": "@dataset.name",
```
robkooper, line 388: can we wrap that in HTML as well, so I can't name my dataset `"; script("popup evil stuff");`

app/views/dataset.scala.html
```diff
@@ -382,6 +382,24 @@ <h4>Statistics</h4>
...
+		 "author": @Html(dataset.author.to_jsonld()),
+		 "description": "@dataset.description",
+		 "dateCreated": "@dataset.created.format("MMM dd, yyyy")",
+		 "DigitalDocument": "@dataset.files.mkString(",")",
```
robkooper, line 392: can we use a link to the file page instead, also limit to first 10 or so (in case we have 100,000 files in a dataset).

app/views/dataset.scala.html
```diff
@@ -382,6 +382,24 @@ <h4>Statistics</h4>
     <script src="@routes.Assets.at("javascripts/deleteUtils.js")" type="text/javascript"></script>
     <script src="@routes.Assets.at("javascripts/people.js")" type="text/javascript"></script>
     <script src="@routes.Assets.at("javascripts/creators.js")" type="text/javascript"></script>
+
+    <script type="application/ld+json">{"context": { "@@vocab": "https://schema.org/"},
```
robkooper, line 386:
```suggestion
    <script type="application/ld+json">
      @dataset.to_jsonld().stringyfy()
    </script>
```
then have in each class a json_ld function that will make it jsonld, all that is needed at that last point is to add a \@context

app/models/User.scala
```diff
@@ -108,7 +110,28 @@ case class MiniUser(
...
+                   "familyName" -> JsString(lastName.toString()),
```
robkooper, line 130: not sure, but I think you can do:
```suggestion
                   "familyName" -> lastName,
```

app/models/User.scala
```diff
@@ -108,7 +110,28 @@ case class MiniUser(
...
+                   "email" ->  JsString(email.get.toString()),
```
robkooper, line 131: careful, email.get might return None for example for twitter, and throw an exception.

robkooper, line 131:
```suggestion
                   "email" ->  email.getOrElse("")),
```

app/models/User.scala
MBcode: name: fullName included now

app/views/dataset.scala.html
MBcode: I will go with the number returned by dataset.files and will have to map then to page links

app/views/dataset.scala.html
MBcode: yes moved that whole block in dataset.scala.html to to a Dataset.scala method, leaving only script tags; will test a bit 1st though

app/models/User.scala
MBcode: Going from JsObject(Seq(    to   Json.obj(  does allow for this, but everything that is not a string, still needs a .toString()

app/models/User.scala
MBcode: put that in

MBcode: your next comment helped

app/views/dataset.scala.html
MBcode: The whole block is now wrapped, but still want to check it

MBcode: done in both dataset and file

app/models/User.scala
MBcode: done that day

app/views/dataset.scala.html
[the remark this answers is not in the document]
MBcode: Now put right to correct json types

app/views/dataset.scala.html
MBcode: classes and a few it links to all have to_jsonld methods

app/views/dataset.scala.html
MBcode: Just need to check on this, tried to resolve other conversations, but it would not let me

MBcode: Been trying to find CLOWDER_URL internally so I can: files.map(f => clowder_host + "files/" + f)

MBcode: use that env var to override default, and set in Server global, to use for mapping; works well unless there is a preferred place for Server url

MBcode:
issue #335 has to_jsonld method, and change to Utils.baseURL signature
Just want to know how to change conf to test w/different root, 
to be sure fileID->fileURL's baseURL catches that change

robkooper:
in clowder, create folder called custom, in there create file custom.conf and put int here:
```
application.context="/hello/"
```

MBcode:
> tried that custom.conf
"DigitalDocument" : [ "http://localhost:9000/hello/files/621560008412bae5e0f511b4", "http://localhost:9000/hello/files/6247460f84124a7d5ae32e61" ],
works, now for some other tests of the  Utils.BaseURL  change on 1st arg

app/models/Dataset.scala
```diff
@@ -38,6 +39,55 @@ case class Dataset(
...
+    */
+  def to_jsonld(url: String) : JsValue = { 
+     val so = JsObject(Seq("@vocab" -> JsString("https://schema.org/")))
+     val URLb = url.replaceAll("/$", "") 
+     var pic_id = thumbnail_id.getOrElse("")
+     if (pic_id != "") {
+        pic_id = URLb + pic_id 
+     } else ""
+     val datasetLD = Json.obj(
+           "context" -> so,
```
lmarini, line 71: I was playing around the json-ld playground and realized that this needs to be `@context` to be properly parsed.

app/models/File.scala
```diff
@@ -32,7 +34,33 @@ case class File(
...
   stats: Statistics = new Statistics(),
-  status: String = FileStatus.UNKNOWN.toString) // can't use enums in salat
+  status: String = FileStatus.UNKNOWN.toString) { // can't use enums in salat
+    /**
+    * return File as JsValue in jsonld format
+    */
+     def to_jsonld() : JsValue = {
+     val so = JsObject(Seq("@vocab" -> JsString("https://schema.org/")))
+     val fileLD = Json.obj(
+           "context" -> so,
```
lmarini, line 44: Same as above.

app/controllers/Utils.scala
```diff
@@ -13,12 +13,14 @@ object Utils {
   /**
    * Return base url given a request. This will add http or https to the front, for example
    * https://localhost:9443 will be returned if it is using https.
+   *
+   * Change type from Request[Any]
    */
-  def baseUrl(request: Request[Any], absolute: Boolean = true) = {
+  def baseUrl(request: RequestHeader, absolute: Boolean = true) = {
```
lmarini, line 19: Can you explain this change? Thank you.

app/controllers/Utils.scala
```diff
@@ -13,12 +13,14 @@ object Utils {
   /**
    * Return base url given a request. This will add http or https to the front, for example
    * https://localhost:9443 will be returned if it is using https.
+   *
+   * Change type from Request[Any]
```
lmarini, line 17: Should this be deleted?

app/models/Dataset.scala
```diff
@@ -38,6 +39,55 @@ case class Dataset(
...
+           "identifier" -> id.toString,
+           "name" -> name,
+           "author" -> author.to_jsonld(),
+           "description" -> description,
+           "dateCreated" -> Formatters.iso8601(created), 
+           "DigitalDocument" -> Json.toJson(cap_api_list(files, 10, URLb, "/files/")), 
+           //"Directory" -> Json.toJson(folders), //skip
```
lmarini, line 78: Can remove commented out lines?

lmarini (changes requested):
I'll play a bit more with it in the next couple of days, but I wanted to leave a couple of quick comments. Thanks!

app/models/File.scala
MBcode: tnx! updated

app/views/dataset.scala.html
```diff
@@ -382,6 +385,11 @@ <h4>Statistics</h4>
     <script src="@routes.Assets.at("javascripts/deleteUtils.js")" type="text/javascript"></script>
     <script src="@routes.Assets.at("javascripts/people.js")" type="text/javascript"></script>
     <script src="@routes.Assets.at("javascripts/creators.js")" type="text/javascript"></script>
+
+    <script type="application/ld+json">
+				@Html(Json.prettyPrint(dataset.to_jsonld(Utils.baseUrl(request))))
```
robkooper, line 390: can you reformat this text

app/views/file.scala.html
```diff
@@ -73,6 +75,10 @@
 	<script src="@routes.Assets.at("javascripts/sectionsListProcess.js")" type="text/javascript"></script>
 	<script src="@routes.Assets.at("javascripts/files/datasets.js")" type="text/javascript"></script>
 
+	<script type="application/ld+json">
+        @Html(Json.prettyPrint(file.to_jsonld()))
+     </script>
```
robkooper, line 80: can you reformat this text

app/controllers/Utils.scala
MBcode: views/dataset.scala.html's has the request, that works to get the needed baseURL
@Html(Json.prettyPrint(dataset.to_jsonld(Utils.baseUrl(request))))
That page already has too many args, making sending the other request type not easy,
&I tried a 2nd version of baseUrl, but was suggested that we just keep one method

app/controllers/Utils.scala
MBcode: When it was suggested we only have 1 version of the method, I left that in case any problems showed up later,
but these notes/etc should be enough of a record.  So I deleted it

app/models/Dataset.scala
MBcode: yes

app/models/ServerStartTime.scala
```diff
@@ -2,6 +2,8 @@ package models
 
 import java.util.Date
 
+import java.net._
```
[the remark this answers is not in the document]
MBcode: leftover from old work around, tnx, removed

app/views/dataset.scala.html
MBcode: got rid of tab, &space to align script

app/views/file.scala.html
MBcode: same change as above

app/models/Dataset.scala
MBcode: replaced and tested

app/models/Dataset.scala
```diff
@@ -38,6 +38,32 @@ case class Dataset(
...
+  def to_jsonld(url: String) : JsValue = { 
+     val so = JsObject(Seq("@vocab" -> JsString("https://schema.org/")))
+     val datasetLD = Json.obj(
+              "context" -> so,
+              "identifier" -> id.toString,
+              "name" -> name,
+              "author" -> author.to_jsonld(),
+              "description" -> description,
+              "dateCreated" -> created.toString.format("MMM dd, yyyy"),
+              //"DigitalDocument" -> Json.toJson(files),
```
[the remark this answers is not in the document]
MBcode: removed commented lines

app/models/LicenseData.scala
```diff
@@ -41,5 +44,72 @@ case class LicenseData (
...
     }
-}
 
+    /**
+     * Utility to return a url even if empty, but enough other attributes available to determine it
+     * this is repurposed from:
+     * function updateData(id, imageBase, sourceObject, authorName)  
+     * in updateLicenseInfo.js line:88
+     */
+    def urlViaAttributes() : String = {
```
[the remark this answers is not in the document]
MBcode: Not sure what you mean, this returns a URL-string, for the license
so there isn't really anything to convert

app/models/Dataset.scala
```diff
@@ -38,6 +38,32 @@ case class Dataset(
...
+              "DigitalDocument" -> Json.toJson(files.map(f => url.replaceAll("/$", "") + "/files/" + f)),
```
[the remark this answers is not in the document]
MBcode: my branch has: "DigitalDocument" -> Json.toJson(cap_api_list(files, 10, URLb, "/files/")),
&there is an issue to make this configurable later

app/models/File.scala
```diff
@@ -32,7 +33,33 @@ case class File(
...
+           "identifier" -> id.toString,
+           "name" -> filename,
+           "author" -> author.to_jsonld(),
+           "isBasedOn" -> originalname,
+           "uploadDate" -> uploadDate.toString.format("MMM dd, yyyy"),
```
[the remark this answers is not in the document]
MBcode: only date now in this file uses it: "uploadDate" -> Formatters.iso8601(uploadDate),

Issue #343, opened 2022-04-01
"Select a collection..." dropdown on dataset returns 400


max-zilla:
https://clowder.ncsa.illinois.edu/clowder/datasets/5e398b4f5e0e2b6b85510821

URL that is called when the dropdown is clicked has incorrect parameter mapping: https://clowder.ncsa.illinois.edu/clowder/api/collections/datasetPossibleParents/undefined?when=5e398b4f5e0e2b6b85510821&date=5

(note that ?when= contains the dataset ID)

robkooper:
I wonder if this is the same as https://github.com/clowder-framework/clowder/pull/307/files#diff-f93c5f084201480bb2e10ed9a303a60ea00553725d1d156dae74ca6e78cffcc5

max-zilla:
https://github.com/clowder-framework/clowder/pull/345

PR #345, opened 2022-04-04
Include 'when' parameter on sub-collection select

max-zilla:
Fixes bug where "Select a Collection" Parent Collection dropdown on Collection page was not working due to missing parameter (#343)

This can be a hotfix if others agree.


robkooper:
can you update changelog and I can add it to 1.20.1

robkooper (changes requested):
can you update changelog

PR #348, opened 2022-04-07
Added logic for primary Clowder instance which handles multiple extra…

ddey2:
…ctors problem and multiple indexing issue

## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [x] When possible

## Types of changes

## Checklist:

- [x] I have signed the CLA

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
+
+### Fixed
 - Fixes bug where same extractor shows up multiple times and all Clowder instances index db on reindex [#327](https://github.com/clowder-framework/clowder/issues/327)
 
 ## 1.20.1 - 2022-04-04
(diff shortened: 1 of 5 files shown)


conf/application.conf
```diff
@@ -190,6 +190,7 @@ mongodbURI = "mongodb://127.0.0.1:27017/clowder"
 # All requests to extractors and any other external process is send
 # using rabbitmq, this will setup the connection information.
 clowder.rabbitmq.uri="amqp://guest:guest@localhost:5672/%2f"
+clowder.primary=true
```
robkooper, line 193: this is in the middle of the rabbitmq section, probably better to put it in its own area and add some text what this does (mention that it is only needed in case you use multiple instances of clowder).

CHANGELOG.md
```diff
@@ -5,6 +5,9 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## Unreleased
```
robkooper, line 8:
```suggestion
## Unreleased

### Fixed

```

add  a section header to indicate that this is a fix.

robkooper (changes requested):

conf/application.conf
ddey2: Took care of this.

robkooper (approved):

Issue #351, opened 2022-04-15
Create sitemap.xml for all the dataset pages


MBcode:
We want the new schema.org metadata from issue #335 to be findable by  https://datasetsearch.research.google.com via a sitemap.xml listing those pages

The url for the sitemap could go to a route that generates it on the fly, or it could be done via a chron job and cached (maybe as a cfg option).

Some places have so much to crawl that their main sitemap has links to sub sitemaps of say 1k links each. We will have to allow for this.

The main starting place is deciding if the whole enpoint should be made findable for a crawl, or some subset (eg. space), so this could also end up as a cfg option at some point, down the road.

MBcode:
[Wrote simple py](https://github.com/MBcode/ec/blob/master/clowder/sitemap.py) requests to create a usable sitemap.xml as a start; 
we can iterate as much as ppl like, have [notes](http://geocodes.ddns.net/clowder/m4) on howto do w/in scala
but then it can't be reused in v2

MBcode:
For very large spaces with very large sets of datasets, that don't change all that much (eg. time-series) allow for: sitemap at _dataset_ or _space_ level, after/near-where you set it to be _public_.
   
To have google and others harvest it, we will have to call the space a _schema:Dataset_ but can have another element that shows, that is is actually something more like a _DataCatalog_

So I will sketch up something like the other [mapping](http://geocodes.ddns.net/clowder/map.htm) here. So we can get further comment on any new elements.   
So we can add a new to_jsonld method to implement this mapping underneath the spaces html pages.

UX-wise, alternately we could always list the space, and just have the radio button have: private,public, pubic-w/sitemap

MBcode:
Starting mapping [look like this](http://geocodes.ddns.net/clowder/m06_): spaceLD = Json.obj(
+           "context" -> so,
+           "identifier" -> id.toString,
+           "name" -> name,
+           "description" -> description,
+           "dateCreated" -> Formatters.iso8601(created),
+           "creator" -> Json.toJson(creator)

MBcode:
I will consolidate all the linked notes, into one summary, that we can ok for me to move forward on more

MBcode:
Most of the changes already creeped into the last pr (other class to ld+json scripts), except for actually making the sitemap, that has code, but no comments on; so not sure if I have the ok to finish this?

MBcode:
[...]

PR #389, opened 2022-10-14
vs using sitemap.xml fork branch

MBcode:
## Review Time Estimate

- [ ] Immediately
- [ ] Within one week
- [ x] When possible

## Types of changes

- [ x] New feature (non-breaking change which adds functionality)

## Checklist:

- [ x] I have updated the CHANGELOG.md.
- [ ] I have signed the CLA

### DIFF
--- a/app/controllers/Application.scala
+++ b/app/controllers/Application.scala
@@ -4,6 +4,7 @@ import models.{Event, UUID, UserStatus}
 import play.api.Play.current
 import play.api.mvc.Action
 import play.api.{Logger, Play, Routes}
+import play.api.libs.json._
 import services._
 import util.Formatters.sanitizeHTML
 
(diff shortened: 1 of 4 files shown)


MBcode:
add-sitemap-route branch for [issue351](https://github.com/clowder-framework/clowder/issues/351)

app/controllers/Application.scala
```diff
@@ -84,6 +85,51 @@ class Application @Inject()(files: FileService, collections: CollectionService,
...
+          clowderurl.getHost
+        } else {
+          clowderurl.getHost + ":" + clowderurl.getPort
+        }
+        var resultStr=""
+        val top= """<?xml version="1.0" encoding="UTF-8"?>
+            <urlset xmlns="http://www.sitemaps.org/schemas/sitemap/0.9"> """
+        resultStr = resultStr.concat(top)
+        //though had called the route2get but couldn't change datastruct
+        val d = scala.io.Source.fromURL(clowderurl + "/api/datasets")
```
robkooper, line 109: why not just call the code directly? `datasets.list()`

MBcode: That got lost, will get it in there, tnx

MBcode: doing a getDatasets for User.anaoymous for now
can look at possible caching, variable list lengths in next iteration

lmarini (changes requested):
Currently we return all datasets, including private ones. I think we should only return public ones. We are calling `TreeService.getDatasets()` which returns datasets `Permission.ViewDataset`. We should probably just return public datasets.

PR #429, opened 2023-10-27
Fixed hearthbeat bug 

lmarini:
where if the owner was not set in extractors_info.json clowder would throw and error and the extractor would not get registered.

### DIFF
--- a/version.sh
+++ b/version.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-VERSION="1.19.2"
+VERSION="1.22.1"
 
 sed -i~ "s#^  val version = \".*\"\$#  val version = \"${VERSION}\"#" project/Build.scala
 sed -i~ "s#^  version: .*\$#  version: ${VERSION}#" public/swagger.yml
--- a/version.sh
+++ b/version.sh
@@ -3,6 +3,7 @@
 VERSION="1.22.1"
 
 sed -i~ "s#^  val version = \".*\"\$#  val version = \"${VERSION}\"#" project/Build.scala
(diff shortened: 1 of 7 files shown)


lmarini:
@max-zilla can you test a private extractor that includes an owner in extractor_info.json. Thank you.

version.sh
```diff
@@ -1,9 +1,8 @@
 #!/bin/bash
 
-VERSION="1.22.0"
+VERSION="1.22.1"
 
 sed -i~ "s#^  val version = \".*\"\$#  val version = \"${VERSION}\"#" project/Build.scala
-sed -i~ "s#^version: .*\$#version: \"${VERSION}\"#" citation.cff
```
robkooper, line 6: why not update the citation file?

max-zilla (approved):
all seems to work as expected.

minump:
I was able to test this as well. Ran `ghcr.io/clowder-framework/clowder:PR-429`, deleted a private extractor from UI (http://localhost:8000/extractors/pdfminer-extractor) . Saw that the extractor came back on (name popped up in the list of extractors) after heartbeat time.

Tested. works fine. approving.

lmarini:
> I was able to test this as well. Ran `ghcr.io/clowder-framework/clowder:PR-429`, deleted a private extractor from UI (http://localhost:8000/extractors/pdfminer-extractor) . Saw that the extractor came back on (name popped up in the list of extractors) after heartbeat time.
> 
> Tested. works fine. approving.

Thanks! Next time feel free to add yourself to list of reviewers.

PR #449, opened 2024-04-28
Fix profile xss

tcnichol:
you can change the firstName and lastName to include javascript using the endpoint:

`localhost:9002/api/users/662eacbe89108d20fb4cab22/updateName?firstName=bb&lastName=<script>alert('XSS')</script>`

I am using an imported function. This gets rid of the javascript being called, but then the name is blank on the profile page. It may not be the right method for this, or I may be using it wrong.

### DIFF
--- a/app/api/Datasets.scala
+++ b/app/api/Datasets.scala
@@ -718,7 +718,6 @@ class  Datasets @Inject()(
     } else {
         val foldersContainingFile = folders.findByFileId(file.id).sortBy(_.name)
         Logger.debug("File was already in dataset.")
-        val foldersContainingFile = folders.findByFileId(file.id).sortBy(_.name)
         Logger.debug("Remove file from folders in dataset")
         folders.get(foldersContainingFile).foreach(folder => {
           if (folder.parentDatasetId == dsId){
--- a/app/api/Datasets.scala
+++ b/app/api/Datasets.scala
@@ -716,7 +716,15 @@ class  Datasets @Inject()(
       }
       Logger.debug("----- Adding file to dataset completed")
     } else {
+        val foldersContainingFile = folders.findByFileId(file.id).sortBy(_.name)
         Logger.debug("File was already in dataset.")
+        val foldersContainingFile = folders.findByFileId(file.id).sortBy(_.name)
+        Logger.debug("Remove file from folders in dataset")
+        folders.get(foldersContainingFile).foreach(folder => {
+          if (folder.parentDatasetId == dsId){
+            folders.removeFile(folder.id, fileId)
+          }
+        })
     }
   }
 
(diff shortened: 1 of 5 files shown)


longshuicy:
@lmarini and I looked together on other possible places related to this bug and realize that sometime it's unnecessary for the frontend template to use @Html(). It should just render it as plain text instead of as html tags.

- The only exception I can think of is Customize the "welcome" page. 
- Maybe even for the places that needs "html" tag, we can exclude <script> tag

e.g.
https://github.com/search?q=repo%3Aclowder-framework%2Fclowder+%40html&type=code&p=1

tcnichol:
Replacing 

`@Html(profile.lastName)`

with 

`(profile.lastName)`

still calls the javascript function.

I am going to try writing an escape function that we can use.

app/api/Datasets.scala
```diff
@@ -716,7 +716,15 @@ class  Datasets @Inject()(
       }
       Logger.debug("----- Adding file to dataset completed")
     } else {
+        val foldersContainingFile = folders.findByFileId(file.id).sortBy(_.name)
         Logger.debug("File was already in dataset.")
+        Logger.debug("Remove file from folders in dataset")
+        foldersContainingFile.foreach(folder => {
```
longshuicy, line 722: Any reason this logic is needed here?

app/views/profile.scala.html
```diff
@@ -65,7 +65,7 @@ <h1>@profile.fullName</h1>
             } else {
                 @if(ownProfile){
                     <div id="prf-first-name" class="text-left inline">
-                        <h1 id="first-name-title" class="inline" style="cursor:pointer" title="Click to edit user's first name.">@Html(profile.firstName)</h1>
+                        <h1 id="first-name-title" class="inline" style="cursor:pointer" title="Click to edit user's first name.">@Html(escapeString("<script>alert('XSS')</script>"))</h1>
```
longshuicy, line 68: What happen if you just remove the @Html() altogether? Like we discussed, this way no html tag should be executed

longshuicy: Also shouldn't this be profile.firstName still? Looks like you might've pushed an temporary test to this PR.

longshuicy: Last thing, there might be other places with @Html tag, could you also apply the same fix to those?

docker-compose.yml
```diff
@@ -51,7 +51,7 @@ services:
 
   # main clowder application
   clowder:
-    image: clowder/clowder:${CLOWDER_VERSION:-latest}
+    image: clowder:bugfix
```
longshuicy, line 54: Again this looks like a temp commit, should it revert back to the original image?

longshuicy (changes requested):

lmarini:
Fixed https://github.com/clowder-framework/clowder/pull/459

PR #453, opened 2024-08-23
adding creator to space json returned, adding route for users in space

### LINKED ISSUE #452: add creator to json returned by space api endpoints
Currently when we get spaces through the api, the creator is not included. This would create difficulties for migration from v1 to v2.

tcnichol:
Currently the space json does not include the creator for the api routes. I have added it since this will be needed for migrating spaces.

### DIFF
--- a/app/api/Spaces.scala
+++ b/app/api/Spaces.scala
@@ -90,6 +90,18 @@ class Spaces @Inject()(spaces: SpaceService,
       case None => NotFound("Space not found")
     }
   }
+  def getUsers(id: UUID) = PermissionAction(Permission.ViewSpace, Some(ResourceRef(ResourceRef.space, id))) { implicit request =>
+    spaces.get(id) match {
+      case Some(space) =>
+
+        val usersInSpace = spaces.getUsersInSpace(space.id, None)
+
+        Ok(toJson(usersInSpace.map(userToJson)))
+      case None => {
+        NotFound("Space not found")
+      }
+    }
+  }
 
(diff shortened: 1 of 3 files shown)


app/api/Spaces.scala
```diff
@@ -167,9 +177,17 @@ class Spaces @Inject()(spaces: SpaceService,
     toJson(Map("id" -> space.id.stringify,
       "name" -> space.name,
       "description" -> space.description,
+      "creator" -> space.creator.stringify,
```
longshuicy, line 180: Does space model has a creator field?

app/api/Spaces.scala
```diff
@@ -167,9 +177,17 @@ class Spaces @Inject()(spaces: SpaceService,
...
       "created" -> space.created.toString))
   }
 
+  def userToJson(user: User) = {
+    toJson(Map("id" -> user.id.stringify,
+      "email" -> user.email.get
+    ))
+  }
```
longshuicy, line 188: I don't know scala well enough, but since you are returning the object of user, why don't you cast every field available?

longshuicy (changes requested):
Code looks good but I haven't got time to test it. 
One thing I noticed, you probably also need to add an entry on Changelog.

app/api/Spaces.scala
tcnichol: yes. it does, but it was not returned.

app/api/Spaces.scala
tcnichol: We already create the user in v2 and have the id from v1, so this is enough for migrating.

tcnichol:
I added an entry to changelog, hoping to get this merged and a new release soon so we can work on migration

